PR #1, opened 2022-04-29
Dockerfile

Davenury:


Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN gradle installDist --no-daemon
+
+###
+
+FROM adoptopenjdk:11-jre-hotspot
+WORKDIR /application
+
+COPY --from=builder /home/gradle/src/build/install/PoC .
+
+ENTRYPOINT ["bash", "-c", "/application/bin/PoC"]
```
Rados13, line 17: Why do we use two different docker images?

build.gradle.kts
```diff
@@ -27,4 +27,10 @@ dependencies {
...
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit:$kotlin_version")
-}
\ No newline at end of file
+}
+
+tasks.withType<Jar> {
+    manifest {
+        attributes["Main-Class"] = "com.example.ApplicationKt"
+    }
+}
```
Rados13, line 36: What do these lines?

Davenury: It's setting main class (entry point, where the `main` function is) - tbh I don't know why it's needed but without them, application didn't want to start from jar.

Dockerfile
Davenury: It's kinda "how it should be done" but I don't know if we need them. I saw this approach in courses and, as far as I can guess, it's because of convenience - the first one is gradle image and we're using it for building, because there already is gradle on it. Second one is lighter, as it's only jdk (most probably without gradle). You can think about it this way - I'm building application on my main computer, where I have gradle installed, then I'm copying just jar file on my other computer, that has too low specification for installing gradle on it.

Rados13: I am not sure if second is lighter, because first container is based on Linux Alpine which is the lightest linux distribution, second is without Alpine, so I assume it is based on Ubuntu, if I remember correctly convention in naming docker images. So I would stay in the same Linux distribution if we would need extend this Dockerfile.

Rados13: https://docs.docker.com/develop/develop-images/multistage-build/
This page answer my questions.

Rados13 (approved):

Dockerfile
Davenury: I could also go with just some linux kernel, as we don't use "java" in entrypoint, just the binary. I can check if it still runs after change to some basic image.

PR #2, opened 2022-04-30
History management

Davenury:


src/main/kotlin/com/example/api/Routing.kt
```diff
@@ -0,0 +1,20 @@
...
+    val historyManagementFacade: HistoryManagementFacade by inject()
+
+    // Starting point for a Ktor app:
+    routing {
+        get("/") {
+            call.respondText("Hello World!")
+        }
+    }
+    routing {
+    }
```
Rados13, line 19: Is this second routing necessary?

src/main/kotlin/com/example/infrastructure/InMemoryHistoryManagement.kt
```diff
@@ -0,0 +1,31 @@
...
+                    historyStorage.add(change)
+                    HistoryChangeSuccess
+                }
+            } }
+
+    override fun getLastChange(): Change? =
+        try {
+            historyStorage.last()
+        } catch (ex: java.util.NoSuchElementException) {
+            null
```
Rados13, line 28: I would add printing some logs, about what happens.

src/test/kotlin/com/example/api/HistoryManagementFacadeSpec.kt
```diff
@@ -0,0 +1,26 @@
+package com.example.api
+
+import com.example.domain.*
+import com.example.infrastructure.InMemoryHistoryManagement
+import com.example.utils.DummyConsensusProtocol
+import org.junit.jupiter.api.Test
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+
+class HistoryManagementFacadeSpec {
```
Rados13, line 10: From what is `Spec` suffix?

Rados13 (commented):
Also two additional comments:
- I am not sure if we need in this simple project a dependency injection library,
- If I remember correctly we need, also a PeerSet data class and I would start HistoryManagment class by adding a Peer to PeerSet.

src/main/kotlin/com/example/api/Routing.kt
Davenury: No, it was provided with example code from intellij, but for now I didn't touch this routes, as I don't know what to do with them now (also it's orthogonal to this task 😂)

src/test/kotlin/com/example/api/HistoryManagementFacadeSpec.kt
Davenury: From Specification https://stackoverflow.com/a/40222901

Rados13: I supposed that, but I wanted to make sure and propose that we would change it to better name, but as it is a convention, it can stay as it is.

src/main/kotlin/com/example/api/Routing.kt
```diff
@@ -11,10 +14,11 @@ fun Application.configureRouting() {
 
     // Starting point for a Ktor app:
     routing {
-        get("/") {
-            call.respondText("Hello World!")
+        post("/change") {
+            val change = ChangeDto(call.receive())
+            println("got call: $change")
```
Rados13, line 19: As we have a logger shouldn't we use it here also?

Davenury: That's something I was about to delete and then I got to the status pages and forgot about this line 😅

Rados13 (approved):

PR #3, opened 2022-05-04
Add ratis library for raft protocol

Rados13:
The first version of using RAFT protocol with Ratis library.
Currently, it exposes REST API for a distributed counter. Next steps will be about adding support for different operation that increment and get.


src/main/kotlin/com/example/api/Routing.kt
```diff
@@ -1,16 +1,19 @@
...
 import io.ktor.request.*
 import io.ktor.response.*
 import io.ktor.routing.*
 import org.koin.ktor.ext.inject
 
-fun Application.configureRouting() {
+fun Application.configureRouting(id: Int) {
 
     val historyManagementFacade: HistoryManagementFacade by inject()
+    val raftNode = RaftNode.initialize(id)
```
Davenury, line 15: I think that it would be better done in Application and pass raftNode as an argument to configureRouting. I don't have any strong arguments for it but I feel that routing should be left clean.

src/main/kotlin/com/example/raft/Constants.kt
```diff
@@ -0,0 +1,87 @@
...
+object Constants {
+    lateinit var PEERS: List<RaftPeer>
+    lateinit var PATH: String
+
+    fun rootPath() = Paths.get("").toAbsolutePath().toString()
+
+    init {
+        val properties = Properties()
+        val conf = Paths.get(rootPath(), "/src/main/resources/conf.properties").toString()
+        print(conf)
```
Davenury, line 46: To delete?

src/main/kotlin/com/example/raft/Constants.kt
```diff
@@ -0,0 +1,87 @@
...
+ */
+object Constants {
+    lateinit var PEERS: List<RaftPeer>
+    lateinit var PATH: String
+
+    fun rootPath() = Paths.get("").toAbsolutePath().toString()
+
+    init {
+        val properties = Properties()
+        val conf = Paths.get(rootPath(), "/src/main/resources/conf.properties").toString()
```
Davenury, line 45: I think that there's better way to extract config but that's not something for now

src/main/kotlin/com/example/raft/Constants.kt
```diff
@@ -0,0 +1,87 @@
...
+        try {
+            FileInputStream(conf).use { inputStream ->
+                InputStreamReader(inputStream, StandardCharsets.UTF_8).use { reader ->
+                    BufferedReader(reader).use { bufferedReader ->
+                        properties.load(bufferedReader)
+                    }
+                }
+            }
+        } catch (e: IOException) {
+            throw IllegalStateException("Failed to load $conf", e)
```
Davenury, line 56: Maybe "Failed to load ratis configuration", e?

src/main/kotlin/com/example/raft/Constants.kt
```diff
@@ -0,0 +1,87 @@
...
+        val key = "raft.server.address.list"
+        val addresses = Optional.ofNullable(properties.getProperty(key))
+            .map { s: String ->
+                s.split(
+                    ","
+                ).toTypedArray()
+            }
+            .orElse(null)
+        require(!(addresses == null || addresses.size == 0)) { "Failed to get $key from $conf" }
+        val key1 = "raft.server.root.storage.path"
```
Davenury, line 67: Key and key1 are not quite good variable names 🤔

src/main/kotlin/com/example/raft/Constants.kt
```diff
@@ -0,0 +1,87 @@
...
+        val path = properties.getProperty(key1)
+        PATH = path ?: "./raft-examples/target"
+        val peers: MutableList<RaftPeer> = ArrayList(addresses.size)
+        for (i in addresses.indices) {
+            peers.add(RaftPeer.newBuilder().setId("n$i").setAddress(addresses[i]).build())
+        }
+        PEERS = Collections.unmodifiableList(peers)
+    }
+
+    private val CLUSTER_GROUP_ID = UUID.fromString("02511d47-d67c-49a3-9011-abb3109a44c1")
```
Davenury, line 77: Is it some kind of constant or did you just paste random uuid from some website? I mean that it probably could go to config too

Davenury (commented):
To tak na teraz, po pracy siądę do tego jeszcze

src/main/kotlin/com/example/raft/Constants.kt
Rados13: This file is from the tutorial so I left it as it is, but I can move it to the config file.

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
```diff
@@ -0,0 +1,19 @@
+package com.example.infrastructure
+
+import com.example.domain.Change
+import com.example.domain.HistoryManagement
+import com.example.raft.HistoryRaftNode
+import org.slf4j.LoggerFactory
+
+class RatisHistoryManagement(private val historyRaftNode: HistoryRaftNode) : HistoryManagement(historyRaftNode) {
```
Davenury, line 8: I don't think leaving default implementation in `HistoryManagement` was a good idea. Right now I was a little confused that this HistoryManagement don't have the other method.

src/main/kotlin/com/example/module.kt
```diff
@@ -2,15 +2,15 @@ package com.example
...
 import com.example.domain.HistoryManagement
 import com.example.infrastructure.DummyConsensusProtocol
 import com.example.infrastructure.InMemoryHistoryManagement
-import io.ktor.http.*
 import org.koin.dsl.module
 
 val historyManagementModule = module {
-    single<ConsensusProtocol> { DummyConsensusProtocol }
+    single<ConsensusProtocol<Change, MutableList<Change>>> { DummyConsensusProtocol() }
     single<HistoryManagement> { InMemoryHistoryManagement(get()) }
```
Davenury, line 14: Right here you can use `RatisHistoryManagement`

src/main/kotlin/com/example/raft/CounterRaftNode.kt
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.File
+
+class CounterRaftNode(peerId: Int) :
```
Davenury, line 23: Do we need it now?

src/main/kotlin/com/example/raft/HistoryStateMachine.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import com.example.domain.Change
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.ObjectMapper
+
+typealias History = MutableList<Change>
+
+
+class HistoryStateMachine(override var state: History = mutableListOf()) : StateMachine<History>(state) {
+    val objectMapper = ObjectMapper()
```
Davenury, line 11: you can import object mapper here from some file (I don't remember now) - as far as I remember creating object mapper is expensive operation

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
Rados13: As HistoryManagment is an abstract class I don't see this as very strange, but I could remove `change` function also abstract and repeat it implementation in both classes extending this function.

src/main/kotlin/com/example/raft/CounterRaftNode.kt
Rados13: I think we can use it for some simple testing, but if you prefer to remove it I can remove it.

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
Davenury: Personally, I think it would be better, as it would be more readable (at least for me).

src/main/kotlin/com/example/raft/CounterRaftNode.kt
Davenury: Right now, we would be testing library. We know it's working, so I think there's no point in it. For tests, I would use code that would actually be used. And if we need this in the future, it's in git, so we can get it back pretty easily. So yes, I'd remove it.

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
Rados13: The problem I see with it is that there would be no sense to keep `HistoryManagment` as an abstract class.

src/main/kotlin/com/example/module.kt
Rados13: That could be problematic because to use RatisHistoryManagment we need two nodes.

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
Davenury: It can be interface. Tbh it doesn't matter to me and still this difference is not quite clear to me.

Although if you meant that there's no point in keeping this level of abstraction - I'd still argue that this is needed. If we're ever going to change it (and I think we will) this kind of abstraction will make it easier. Also - with this testing is easier as you can mock interface or even use some dummy implementation.

src/main/kotlin/com/example/module.kt
Davenury: As far as I can see, we need only RaftHistoryNode in constructor. I don't see any problem with this, but I would need to play around with code to be sure. So let it be for now, after merge I'll see if this can be done.

src/main/kotlin/com/example/api/Routing.kt
```diff
@@ -1,23 +1,29 @@
...
         post("/change") {
             val change = ChangeDto(call.receive())
-            historyManagementFacade.change(change)
-            call.respond(HttpStatusCode.OK)
+            val result = historyManagement.change(change.toChange())
+            call.respond(result.toString())
+        }
+        get("/change") {
+            val result = historyManagement.getLastChange()
+            call.respond(result?.let { objectMapper.writeValueAsString(it) } ?: "Error")
```
Davenury, line 24: I'd stick with one kind of response. If it's text, let it be, if it's objects - let it be. I remember from e-Stella that it was kind of hard to keep it concise and there were a lot of problems with that from front-end side (these problems will also occur when we'll be using rest calls on backend).

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
Rados13: My point is that in previous talk on discord your point was that we need this as abstract class with, because each class which extend/implement it have to use Consensus protocol variable. So I leave this way and add implementation for a method which for all class will be the same. But you don't like it this way and now you suggest to make interface.

src/main/kotlin/com/example/module.kt
Rados13: It is required because raft as consensus protocol can't work with one node XD

src/main/kotlin/com/example/infrastructure/RatisHistoryManagement.kt
Davenury: Yeah, I remember. I just didn't know (and still don't) if this consensus protocol is needed as field of this class. But it's ok to leave it as it is, even with this default implementation.

src/main/kotlin/com/example/module.kt
Davenury: What I meant is that you need just one node to instantiate this class. But maybe I didn't see something, because I'm doing this review on phone 😅

src/main/kotlin/com/example/domain/Change.kt
```diff
@@ -24,16 +25,23 @@ data class ChangeDto(
 @JsonIgnoreProperties(ignoreUnknown = true)
 sealed class Change(
     val operation: Operation
-)
+) {
+    companion object {
+        fun fromJson(json: String): Change = Gson()
```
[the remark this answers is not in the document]
Davenury: bump

src/main/kotlin/com/example/raft/HistoryRaftNode.kt
```diff
@@ -26,8 +22,8 @@ class HistoryRaftNode(peerId: Int) :
         val msg = HistoryStateMachine.OperationType.STATE.toString()
         val result = queryData(msg)
         return try {
-            Gson()
-                .fromJson(result, mutableListOf<LinkedTreeMap<String, String>>().javaClass)
+            objectMapper
+                .readValue(result, mutableListOf<LinkedHashMap<String, String>>().javaClass)
```
Davenury, line 26: I've always used (and seen in usage) `::class.java` but I don't know the difference, so that's just as a curiosity

Davenury (approved):

PR #8, opened 2022-05-11
Dynamic group

Rados13:
In this PR three I tried to implement two features: rejecting consensus propose and dynamic adding node to raft group. I failed in both tasks. I will summarize in short results.

#### Rejecting consensus propose
Ratis library API doesn't allow each node to decide, if they accept or reject consensus proposition. I also try to hack it, by changing implementation of making change of state machine state. I thought that change is made in each followers the same way it is made in leader. But after many tries it comes out that change is propageted different way or leader doesn't propagate change, but the whole new state to followers.

#### Dynamic adding node to raft group
Also Ratis library API doesn't allow to change configuration of raft server. Tries of hacking it leads to strange bugs like infinite loop of sending same message. The furthest progress was when it looked like the connection between the configuration node and dynamic node was made, but this connection wasn't stable enough to allow us to make any operation.

#### Summary
These experiences raise the question, of whether we shouldn't look for another library for consensus protocol.

### DIFF
--- a/src/main/kotlin/com/example/Application.kt
+++ b/src/main/kotlin/com/example/Application.kt
@@ -1,10 +1,9 @@
 package com.example
 
 import com.example.api.configureRouting
-import com.example.domain.ErrorMessage
-import com.example.domain.MissingParameterException
-import com.example.domain.UnknownOperationException
+import com.example.domain.*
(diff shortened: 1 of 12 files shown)


Rados13:
Not needed anymore

PR #9, opened 2022-05-12
Protocol - happy path + errors when cohorts aren't responding

Davenury:
no leader faults


build.gradle.kts
```diff
@@ -27,8 +27,11 @@ dependencies {
     //ktor
     implementation("io.ktor:ktor-server:$ktor_version")
     implementation("io.ktor:ktor-server-netty:$ktor_version")
-    implementation("io.ktor:ktor-serialization:$ktor_version")
+    implementation("io.ktor:ktor-jackson:$ktor_version")
```
Rados13, line 30: Why did you change that, as a few lines lower, we import object mapper Jackson?

src/main/kotlin/com/example/api/ProtocolRouting.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.ktor.http.*
+import io.ktor.request.*
+import io.ktor.response.*
+import io.ktor.routing.*
+
+fun Application.protocolRouting(protocol: GPACProtocol, otherPeers: List<String>) {
+
+    routing {
+        post("/create_change") {
+            val change = ChangeDto(call.receive())
```
Rados13, line 15: I would make some convention and stay one way, `call.receive<ChangeDto>` or `ChangeDto(call.receive())`.

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -0,0 +1,135 @@
...
+
+data class State(val ballotNumber: Int, val initVal: Accept, val acceptNum: Int, val acceptVal: Accept?, val decision: Boolean) {
+    companion object {
+        fun defaultState() =
+            State(0, Accept.ABORT, 0, null, false)
+
+        fun resetStateAfterTransaction(prevState: State) =
+            State(prevState.ballotNumber, Accept.ABORT, 0, null, false)
+    }
+}
```
Rados13, line 18: I would move it GPACProtocol implementation or GPACProtocol interface as it shouldn't be used outside the context of GPAC and the name of this data class is very generic.

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -0,0 +1,135 @@
...
+
+            // TODO: lock?
+
+            return Agreed(state.ballotNumber, acceptVal)
+        }
+        throw NotElectingYou(this.state.ballotNumber)
+    }
+
+    override fun handleApply(message: Apply) {
+        this.state = this.state.copy(decision = true, acceptVal = Accept.COMMIT)
```
Rados13, line 60: When we abort a transaction acceptVal wouldn't be `Accept.ABORT`?

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -0,0 +1,135 @@
...
+        historyManagement.change(message.change.toChange())
+        this.state = State.resetStateAfterTransaction(this.state)
```
Rados13, line 62: Why do you reset state after a transaction?

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -0,0 +1,135 @@
...
+                        append("Content-Type", "application/json")
+                        append("Accept", "application/json")
+                    }
+                    body = ElectMe(this@GPACProtocolImpl.state.ballotNumber, change)
+                }
+            } catch (e: Exception) {
+                logger.error("Peer $it responded with exception: $e - election")
+                null
+            }
+        }
```
Rados13, line 101: Is this httpClient blocking? I would use some async/await/future, because IMO it will be accurate to use this here.

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    private suspend fun getAgreedResponses(change: ChangeDto, otherPeers: List<String>, acceptVal: Accept): List<Agreed> =
+        otherPeers.mapNotNull {
+            try {
+                httpClient.post<Agreed>("http://$it/ft-agree") {
+                    contentType(ContentType.Application.Json)
+                    accept(ContentType.Application.Json)
+                    body = Agree(this@GPACProtocolImpl.state.ballotNumber, acceptVal, change)
```
Rados13, line 109: For what purpose is this `@` here?

Rados13 (changes requested):

src/main/kotlin/com/example/api/ProtocolRouting.kt
Davenury: There's difference between those two. As `call.receive` is just a map, it was more natural to create `ChangeDto` with it as a parameter. As for the other way, it will try to serialize received map to this class - ChangeDto has `properties` map as its field, not the things we're sending.

build.gradle.kts
Davenury: It's more about ktor integration with jackson than jackson itself. Up to this PR we've used kotlinx serialization to serialize requests and it's not always perfect.

src/main/kotlin/com/example/domain/GPACProtocol.kt
Davenury: Ok I thought that apply means that we accept and commit this change - I'll change it

src/main/kotlin/com/example/domain/GPACProtocol.kt
Davenury: Originally, there's a line in article that each transaction has its own state. For now I just reset it, as I'm not convinced how can we keep track of the list of transactions without transaction id.

src/main/kotlin/com/example/domain/GPACProtocol.kt
Davenury: It is blocking but I don't see, why async / await would be better?

src/main/kotlin/com/example/domain/GPACProtocol.kt
Davenury: TBH I'm not quite sure. I think about it like "`this` doesn't relate to the block of code you're in (the httpClient.post block in this case), but to the class you're in (GPACProtocolImpl in this case)".

src/main/kotlin/com/example/domain/GPACProtocol.kt
Rados13: Ok, I was thinking about the state as about transaction history. I would change this variable name maybe to pacState.

src/main/kotlin/com/example/domain/GPACProtocol.kt
Rados13: For optimization reasons, e.g. when none of the peers will respond and you will wait for a timeout when it is blocking, you will be waiting `peers_number * timeout`, in async you will wait something closer to `timeout`.

src/main/kotlin/com/example/domain/GPACProtocol.kt
Rados13: Ok

build.gradle.kts
Rados13: Ahh JVM and it's serialization libraries 😮‍💨

Davenury: Ik if we wrote this in Haskel, there would be one (or not ans we would need to write it by ourselves) 😂

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -22,65 +24,107 @@ interface GPACProtocol {
...
+            defaultTransaction.acceptVal,
+            defaultTransaction.decision
+        )
     }
 
     override fun handleAgree(message: Agree): Agreed {
-        if(checkBallotNumber(message.ballotNumber)) {
-            val acceptVal = if(historyManagement.canBeBuild(message.change.toChange())) Accept.COMMIT else Accept.ABORT
-            this.state = this.state.copy(ballotNumber = message.ballotNumber, acceptVal = acceptVal)
+        if (checkBallotNumber(message.ballotNumber)) {
```
Rados13, line 68: I would check the negation of condition and throw the error earlier, it will remove one nesting.

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -22,65 +24,107 @@ interface GPACProtocol {
...
+        throw NotElectingYou(myBallotNumber)
     }
 
     override fun handleApply(message: Apply) {
-        this.state = this.state.copy(decision = true, acceptVal = Accept.COMMIT)
-        historyManagement.change(message.change.toChange())
-        this.state = State.resetStateAfterTransaction(this.state)
+        this.transactions[message.ballotNumber] =
+            this.transactions[message.ballotNumber]?.copy(decision = true, acceptVal = Accept.COMMIT, ended = true)
+                ?: throw java.lang.IllegalStateException("Got apply for transaction that isn't in transactions map: ${message.ballotNumber}")
```
Rados13, line 89: Are you sure that this should be that exception, as the earlier Exception is without `java.lang`?

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -22,65 +24,107 @@ interface GPACProtocol {
...
+
+        if (message.acceptVal == Accept.COMMIT) {
```
Rados13, line 91: Parentheses are unnecessary

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -22,65 +24,107 @@ interface GPACProtocol {
...
+            if (!historyManagement.canBeBuild(change.toChange())) throw HistoryCannotBeBuildException()
+            this.transactions[myBallotNumber] = Transaction.defaultTransaction(ballotNumber = myBallotNumber, initVal = Accept.COMMIT)
             electResponses = getElectedYouResponses(change, otherPeers)
             tries++
-        } while (electResponses.size < otherPeers.size / 2 + 1 && tries < maxLeaderElectionTries)
+            if (electResponses.size <= otherPeers.size / 2 && tries < maxLeaderElectionTries) {
+                myBallotNumber++
+            } else {
+                break
+            }
```
Rados13, line 108: You can add if condition inside while, and on the end of the loop always do incrementation.

src/test/kotlin/com/example/domain/GPACProtocolSpec.kt
```diff
@@ -41,22 +46,36 @@ class GPACProtocolSpec {
...
+        expectThat(subject.getTransaction(100)!!.ballotNumber).isEqualTo(100)
+        expectThat(subject.getBallotNumber()).isEqualTo(100)
+    }
+
+    @Test
+    fun `should throw IllegalState when trying to agree on transaction that wasn't in elect state`() {
+        val message = Agree(100, Accept.COMMIT, changeDto)
+
+        println(subject.getTransaction(100))
+        println(subject.getBallotNumber())
```
Rados13, line 74: Do we still need this?

Rados13 (changes requested):

src/main/kotlin/com/example/Application.kt
```diff
@@ -76,4 +76,4 @@ fun getIdAndOffset(args: Array<String>): NodeIdAndPortOffset {
 }
 
 fun getOtherPeers(peersAddresses: List<String>, nodeId: Int): List<String>
-    = peersAddresses.filterNot { it.split(":").last().toInt() == 8080 + nodeId }
+    = peersAddresses.filterNot { it.contains("peer$nodeId") || it.contains("808$nodeId") }
```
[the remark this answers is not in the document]
Davenury: Go BIG or go home 😄

src/main/kotlin/com/example/domain/GPACProtocol.kt
Davenury: I've heard that you should always add parentheses because in the future it'll be easier to add something inside if (also it's more readable). - And yeah, it was JVM senior, so I think he knows what he's talking about 😄

src/test/kotlin/com/example/domain/GPACProtocolSpec.kt
Davenury: Of course, without them nothing works 😄

src/main/kotlin/com/example/domain/GPACProtocol.kt
Rados13: I don't think that paranthesses are more readable but, if you like it more this way ok, but according to kotlin coding convention paranthesses are required only for multiline statement.
https://kotlinlang.org/docs/coding-conventions.html#control-flow-statements

Rados13 (approved):

PR #10, opened 2022-05-22
events listeners

Davenury:


src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -112,14 +121,21 @@ class GPACProtocolImpl(
             myBallotNumber++
         } while (electResponses.size <= otherPeers.size / 2 && tries < maxLeaderElectionTries)
 
+        myBallotNumber--
```
Rados13, line 124: Why do we have this line?

Rados13: Yes, I think that solution is much better to increment at the start of the do-while loop.

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -111,18 +111,19 @@ class GPACProtocolImpl(
...
-        addons[TestAddon.OnSendingElect]?.invoke(Transaction(this.myBallotNumber, Accept.COMMIT))
+        if(!historyManagement.canBeBuild(change.toChange())) {
+            throw HistoryCannotBeBuildException()
+        }
 
         do {
+            myBallotNumber++
             if (!historyManagement.canBeBuild(change.toChange())) throw HistoryCannotBeBuildException()
             this.transactions[myBallotNumber] = Transaction(ballotNumber = myBallotNumber, initVal = Accept.COMMIT)
+            addons[TestAddon.OnSendingElect]?.invoke(this.transactions[myBallotNumber])
```
Rados13, line 122: I think we will also need a callback also for the state when we successfully became leader.

PR #14, opened 2022-06-08
Leader fail

Davenury:


src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -122,20 +106,70 @@ class GPACProtocolImpl(
...
+        this.transaction = this.transaction.copy(acceptVal = acceptVal)
+
+        signal(TestAddon.BeforeSendingAgree, this.transaction)
+
+        val agreedResponses = ftAgreePhase(change, acceptVal)
+
+        signal(TestAddon.BeforeSendingApply, this.transaction)
+
+        val applyResponses = applyPhase(change, acceptVal)
+    }
```
Rados13, line 166: As each phase has got the same steps but it has got a little different state, this looks similar to a state machine and I think that refactoring it to a state machine can improve readability a lot.

Rados13 (changes requested):

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -122,20 +106,70 @@ class GPACProtocolImpl(
...
     ) {
+        var electResponses = electMePhase(change) { responses -> superMajority(responses) }
+
+        // TODO - check for decision true
+        val messageWithDecision = electResponses.flatten().find { it.decision }
+        if (messageWithDecision != null) {
+            // end of protocol, this cohort only needs to send apply once more to all the other peers
+            val applyMessages = applyPhase(change, messageWithDecision.acceptVal!!)
+            return
+        }
```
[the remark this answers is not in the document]
Davenury: Continue with protocol

src/main/kotlin/com/example/domain/GPACProtocol.kt
Davenury: For today no, for the future - of course 😄

src/main/kotlin/com/example/raft/HistoryStateMachine.kt
```diff
@@ -3,32 +3,37 @@ package com.example.raft
...
-        } catch (e: JsonMappingException) {
-            e.message
-        }
-
+            try {
+                val map = objectMapper.readValue(operation, HashMap<String, Any>().javaClass)
+                val changeString = objectMapper.writeValueAsString(map["change"])
+                val change: Change = Change.fromJson(changeString)!!
+                val acceptNum = map["acceptNum"] as Int?
+                state.add(ChangeWithAcceptNum(change, acceptNum))
```
Rados13, line 23: Create function `ChangeWithAcceptNum.fromJson`.

src/main/resources/single_peerset_application.conf
```diff
@@ -32,5 +32,5 @@ peers {
 }
 
 protocol {
-    leaderFailTimeoutInSecs = 60
+    leaderFailTimeoutInSecs = 3
```
Rados13, line 35: It's quite a big change of timeout 🤣

src/test/kotlin/com/example/api/IntegrationTest.kt
```diff
@@ -129,6 +135,131 @@ class IntegrationTest {
             listOf(app1, app2, app3).forEach { app -> app.cancel(CancellationException("Test is over")) }
         }
 
+    @Test
+    fun `should be able to execute transaction even if leader fails after first ft-agree`() {
+
+        deleteRaftHistories()
```
Rados13, line 141: Could we move this line to `@BeforeEach`?

src/test/kotlin/com/example/api/IntegrationTest.kt
```diff
@@ -129,6 +135,131 @@ class IntegrationTest {
...
+    fun `should be able to execute transaction even if leader fails after first ft-agree`() {
+
+        deleteRaftHistories()
+
+        runBlocking {
+
+            lateinit var app1: Job
+
+            val firstLeaderAction: suspend (Transaction?) -> Unit = {
+                val url = "http://localhost:8082/ft-agree"
```
Rados13, line 148: Could we move these pairs localhost and port to variables like `peer1` and `peer2`?

src/test/kotlin/com/example/api/IntegrationTest.kt
```diff
@@ -129,6 +135,131 @@ class IntegrationTest {
...
+            }
+
+            val map = objectMapper.readValue(response, HashMap<String, Any>().javaClass)
+            val changeString = objectMapper.writeValueAsString(map["change"])
+            val change: Change = Change.fromJson(changeString)!!
+            expectThat(change).isEqualTo(AddUserChange("userName"))
+
+            listOf(app1, app2, app3)
+                .forEach { app -> app.cancel(CancellationException()) }
+        }
```
Rados13, line 190: What do you think about adding a comment that there we check if a transaction was processed after leader failure?

Rados13 (changes requested):
I think further tests should be focused on failures inside shards e.g: we have two shards each with 3 or 5 nodes and what happens when all or majority of nodes from one shard fail before the leader sends ft-agree/apply etc. Because I have a feeling that there are some edge cases regarding shards, which we don't handle properly.

src/main/kotlin/com/example/api/Routing.kt
```diff
@@ -17,14 +17,19 @@ fun Application.configureSampleRouting(historyManagement: HistoryManagement) {
         route("/change") {
             post {
                 val change = ChangeDto(call.receive())
-                val result = historyManagement.change(change.toChange())
+                val result = historyManagement.change(change.toChange(), 1)
```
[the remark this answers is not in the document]
Davenury: The whole purpose of this endpoint was to test Ratis. IMO I would get rid of it.

src/main/kotlin/com/example/domain/ProtocolClient.kt
```diff
@@ -52,6 +67,20 @@ class ProtocolClientImpl : ProtocolClient {
...
+                    } catch (e: ClientRequestException) {
+                        logger.error(errorMessage(it, e))
+                        if (e.response.status.value == 422) {
+                            val value = e.response.content.readUTF8Line()?.let {
+                                return@let Regex("[0-9]+").findAll(it)
+                                    .map(MatchResult::value)
+                                    .toList()
+                                    .map { it.toInt() }
+                            }?.get(0) ?: 0
+                            acc = aggregateErrors(acc, value)
```
Rados13, line 79: Create a function for that or add some description about what happened here.

src/main/kotlin/com/example/domain/ProtocolClient.kt
```diff
@@ -52,6 +67,20 @@ class ProtocolClientImpl : ProtocolClient {
...
+                            null
+                        } else {
+                            null
+                        }
```
Rados13, line 83: You probably can move the null outside the if-else.

src/main/kotlin/com/example/infrastructure/ProtocolTimerImpl.kt
```diff
@@ -5,16 +5,23 @@ import kotlinx.coroutines.GlobalScope
...
-    override fun startCounting(action: suspend () -> Unit) {
+    companion object {
+        private val randomGenerator = Random()
+    }
+
+    override suspend fun startCounting(action: suspend () -> Unit) {
+        cancelCounting()
         task = GlobalScope.launch {
-            delay((delay * 1000).toLong())
+            delay((delay + randomGenerator.nextLong() % backoffBound) * 1000)
```
Rados13, line 24: I assume that 1000 is, because of moving from seconds to milliseconds. Make it variable with a good name.

src/main/kotlin/com/example/utils.kt
```diff
@@ -1,12 +1,15 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import io.ktor.client.*
 import io.ktor.client.engine.okhttp.*
 import io.ktor.client.features.*
 import io.ktor.client.features.json.*
 
-val objectMapper = jacksonObjectMapper()
+val objectMapper: ObjectMapper = jacksonObjectMapper().
+        configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true)
```
Rados13, line 12: What does this configuration change?

src/test/kotlin/com/example/api/SinglePeersetIntegrationTest.kt
```diff
@@ -212,59 +175,64 @@ class IntegrationTest {
...
         // only one change and this change shouldn't be applied for 8082 two times
         expect {
             that(values.size).isEqualTo(1)
-            that(values[0]).isEqualTo(ChangeWithAcceptNum(AddUserChange("userName"), 1))
+            that(values[0]).isEqualTo(ChangeWithAcceptNum(AddGroupChange("name"), 1))
         }
-
-
-        listOf(app1, app2, app3).forEach { app -> app.cancel(CancellationException("Test is over")) }
+        println("here")
```
Rados13, line 225: Do we need all of these `println`?

Rados13 (changes requested):

src/main/kotlin/com/example/api/Routing.kt
```diff
@@ -15,21 +17,29 @@ fun Application.configureSampleRouting(historyManagement: HistoryManagement) {
     // Starting point for a Ktor app:
     routing {
         route("/change") {
-            post {
-                val change = ChangeDto(call.receive())
-                val result = historyManagement.change(change.toChange(), 1)
-                call.respond(result.toString())
-            }
```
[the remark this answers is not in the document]
Davenury: It's an endpoint which gives you access directly to Ratis, without our protocol. Since it's not needed, I removed it

src/main/kotlin/com/example/config.kt
```diff
@@ -1,14 +1,25 @@
...
-    ConfigLoader().loadConfigOrThrow<Config>("/${System.getenv("CONFIG_FILE") ?: "application.conf"}")
\ No newline at end of file
+fun loadConfig(overrides: Map<String, Any> = emptyMap()) =
+    ConfigLoaderBuilder
+        .default()
+        .addSource(MapPropertySource(overrides))
+        .addResourceSource("/${System.getenv("CONFIG_FILE") ?: "application.conf"}")
+        .build()
+        .loadConfigOrThrow<Config>()
+        .also { println(it) }
```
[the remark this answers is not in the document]
Davenury: Considering it logs out whole config, it's not something I would leave for production 😂 tbh I must have forgotten about this one

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -204,19 +223,38 @@ class GPACProtocolImpl(
             if (!historyManagement.canBeBuild(change.toChange())) throw HistoryCannotBeBuildException()
             this.transaction = transaction ?: Transaction(ballotNumber = myBallotNumber, initVal = Accept.COMMIT)
             signal(TestAddon.BeforeSendingElect, this.transaction)
-            electResponses = getElectedYouResponses(change, otherPeers)
+            logger.info("$me - sending ballot number: $myBallotNumber")
+            val (responses, maxBallotNumber) = getElectedYouResponses(change, otherPeers, acceptNum)
+            electResponses = responses
```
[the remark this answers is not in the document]
Davenury: Just for the deconstruction but I'll look how to use it better.

src/main/kotlin/com/example/domain/ProtocolClient.kt
Davenury: > When I wrote this code, only I and God knew what was happening here. Now only God knows.

src/main/kotlin/com/example/domain/GPACProtocol.kt
```diff
@@ -83,21 +81,23 @@ class GPACProtocolImpl(
         signal(TestAddon.OnHandlingAgreeBegin, transaction)
 
         if (!checkBallotNumber(message.ballotNumber)) {
-            throw NotElectingYou(myBallotNumber)
+            throw NotElectingYou(myBallotNumber, message.ballotNumber)
```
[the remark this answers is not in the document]
Davenury: I'll change it to `NotValidLeader` - should be better

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Davenury: https://stackoverflow.com/a/39273672 -> I think that's the way you do it in Kotlin,

src/main/kotlin/com/example/utils.kt
Davenury: TBH, I don't remember. I guess that I had some problem with object mapping and this option either helped or not and I forgot to delete it. Anyway, it doesn't do any harm for now, so I'd leave it be.

src/test/kotlin/com/example/api/SinglePeersetIntegrationTest.kt
Davenury: yeah, without them tests aren't passing 😄

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Rados13: I am unsure if making the same exception for situations when you are not elected as leader and you should be elected as leader but, before you send ft-agree the peer has incremented ballot number is a good idea.

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Rados13: Ok, but my question was more about that `electResponse` isn't changed in this function, so we could name it this way in deconstruction and don't add an additional variable.

src/test/kotlin/com/example/api/SinglePeersetIntegrationTest.kt
Rados13: Could we use the here logger?

src/main/kotlin/com/example/domain/ProtocolClient.kt
```diff
@@ -68,6 +68,9 @@ class ProtocolClientImpl : ProtocolClient {
                             body = requestBody!!
                         }
                     } catch (e: ClientRequestException) {
+                        // since we're updating ballot number in electing phase, this mechanism lets us
+                        // get any aggregation from all responses from "Not electing you" response, so we can get
```
Rados13, line 72: We don't have now NotElectYou exception but I would add it again 😄

src/test/kotlin/com/example/api/SinglePeersetIntegrationTest.kt
Davenury: it was just a joke, I believe that I got rid of them

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Davenury: we can't. When I tried to do that, variable defined above while is other variable and then it's "not initialized" after while.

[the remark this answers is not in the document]
Rados13: Ok so I would add 
```kotlin
if(superFunction(electResponses) return electResponses
```
And after a while stay only with the body of if.

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Davenury: It's still the same thing. In either case you're not a valid leader.

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Davenury: I don't get it

src/main/kotlin/com/example/domain/GPACProtocol.kt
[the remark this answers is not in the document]
Davenury: `Not a valid leader` in ft-agree phase, `Not Electing You` in elect phase

Rados13 (approved):

PR #16, opened 2022-06-26
Set up GitHub Actions

kjarosh:
This change will allow building and testing the application each time it's pushed to origin


.github/workflows/gradle.yml
```diff
@@ -0,0 +1,15 @@
+name: Gradle
+
+on: [ push, pull_request ]
```
Davenury, line 3: In one of old repos we have sth like this:
```yaml
# Controls when the action will run. 
on:
  # Triggers the workflow on push or pull request events but only for the main branch
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
```
I don't know if there's any difference, but right now I don't see checks 😄

kjarosh: Well, yeah, I've tested this configuration and it works. I suppose the checks do not show because I'm an external contributor and you would have to allow for Actions?

Anyways, here's the test: https://github.com/kjarosh/Unifying-Consensus-And-Atomic-Commitment-PoC/pull/2

So the difference between code in this PR and your example is that your example constrains Actions to run on `main` only — I can change to that if you prefer :smiley:

Davenury: I don't see any reason for not having checks for other PRs than to main - can stay for me.

If I remember correctly, the only thing required to run action is to have a workflow yaml file. We can merge it and (if for any reason it won't work) we can look closer to it but it can be that it's not a part of repository yet, so Github doesn't read it.

For me it's ok, you can merge, when you're ready 😄

PR #17, opened 2022-07-05
first multiple peerset test

Davenury:


src/main/kotlin/com/example/raft/Constants.kt
```diff
@@ -24,26 +24,17 @@ import java.util.*
  * limitations under the License.
  */
 
+class RaftConstants(
```
Rados13, line 27: I think it now can have the name `RaftConfiguration` or something like that.

PR #20, opened 2022-07-13
Consensus

Davenury:


src/test/kotlin/com/example/consensus/ConsensusSpec.kt
```diff
@@ -0,0 +1,86 @@
...
+        val peer5 = GlobalScope.launch(Dispatchers.IO) { startApplication(arrayOf("5", "1")) }
+
+        delay(5000)
+
+        // when: peer1 executed change
+        expectCatching {
+            executeChange("$peer1/consensus/create_change")
+        }.isSuccess()
+
+        val changes = askForChanges("http://localhost:8083")
```
Rados13, line 44: Could we move `http://localhost:8083` to variable with proper name?

Rados13 (dismissed):

PR #21, opened 2022-07-18
Enable tests in build and get rid of delays

kjarosh:


src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
```diff
@@ -157,14 +151,22 @@ class SinglePeersetIntegrationTest {
...
-                throw RuntimeException()
+                throw RuntimeException("Stop")
             }
             val firstLeaderCallbacks: Map<TestAddon, suspend (ProtocolTestInformation) -> Unit> = mapOf(
                 TestAddon.BeforeSendingAgree to firstLeaderAction
             )
+            val afterHandlingApply: suspend (ProtocolTestInformation) -> Unit = {
+                phaser.arriveAndAwaitAdvance()
+            }
+            val peer2Callbacks: Map<TestAddon, suspend (ProtocolTestInformation) -> Unit> = mapOf(
```
Rados13, line 162: We have defined typealias for `suspend (protocolTestInformation: ProtocolTestInformation) -> Unit`

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
```diff
@@ -235,10 +234,17 @@ class SinglePeersetIntegrationTest {
             TestAddon.BeforeSendingApply to firstLeaderAction
         )
 
+        val peer2Action: suspend (ProtocolTestInformation) -> Unit = {
+            phaser.arriveAndAwaitAdvance()
+        }
```
Rados13, line 239: Is this name correct, as we refer to it in peer3Callbacks?

Rados13 (approved):

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
kjarosh: Replaced all references to this type in tests

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
kjarosh: Nope, it was a typo, now it should be correct

Rados13 (approved):
One test fails. This is a test with delays so maybe we could skip tests that are using delays until we rewrite them this way that they will not use delays?

kjarosh:
Yup, I've disabled tests with delays

Rados13 (approved):

PR #29, opened 2022-07-30
Package rename

Davenury:


src/main/kotlin/github/davenury/ucac/Application.kt
```diff
@@ -1,16 +1,15 @@
-package com.example
+package github.davenury.ucac
```
kjarosh, line 1: `com.github.davenury.ucac` would be the proper package

Rados13 (approved):

PR #30, opened 2022-08-02
review fixes

Davenury:


kjarosh:
I believe there's an error in commit history: probably a wrong rebase?

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -8,56 +8,54 @@ import kotlin.random.Random
...
+        var currentApp = 0
+        apps = MutableList(numberOfPeersets) { peersetId ->
+            MutableList(numberOfPeersInPeersets[peersetId]) { peerId ->
+                currentApp++
+                createApplication(
+                    arrayOf("${peerId + 1}", "${peersetId + 1}"),
+                    actions[currentApp] ?: emptyMap(),
+                    eventListeners[currentApp] ?: emptyList(),
+                    ratisConfigOverrides + (configOverrides[peerId] ?: emptyMap()),
+                    TestApplicationMode(peerId + 1, peersetId + 1)
```
Rados13, line 37: As I understand add one to `peerId` and `peersetId` because we start iterating from 1, why do we do that to ourselves? I would change that we start iterating like all programmers from 0 😄.

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -8,56 +8,54 @@ import kotlin.random.Random
...
                 )
             }
-            apps.add(peersetApps)
         }
 
         // start and address discovery
-        apps.flatten().forEach { app -> app.startNonblocking() }
+        apps.flatten().forEachIndexed { index, app -> if (index + 1 !in appsToExclude) app.startNonblocking() }
```
Rados13, line 43: What do you think about this change?
```suggestion
        apps
        .flatten()
        .filterIndexed { index, _app -> !appsToExclude.contains(index+1) }
        .forEach{ it.startNonblocking() }

```

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -8,56 +8,54 @@ import kotlin.random.Random
...
         peers =
             apps.flatten()
                 .asSequence()
-                .withIndex()
-                .groupBy{ it.value.getPeersetId() }
+                .mapIndexed {index, it -> Pair(it, if (index + 1 in appsToExclude) "localhost:0" else "localhost:${it.getBoundPort()}") }
```
Rados13, line 47: What do you think about this change?
```suggestion
                .mapIndexed {index, it -> 
                val address = if (index + 1 in appsToExclude) "localhost:0" else "localhost:${it.getBoundPort()}"
                Pair(it, address) 
                }

```

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -8,56 +8,54 @@ import kotlin.random.Random
...
 
-        apps.flatten().forEachIndexed { index, application ->
-            val filteredPeers = peers.map { peerset -> peerset.filter { it != "localhost:${application.getBoundPort()}" } }
-            application.setOtherPeers(filteredPeers)
+        apps.flatten().zip(peers.flatten()).forEachIndexed { index, (app, peer) ->
+            if (index + 1 !in appsToExclude) {
+                app.setOtherPeers(
+                    peers.map { it.filterNot { it == peer } }
+                )
+            }
```
Rados13, line 58: What do you think about using `filterIndexed` first and then `forEach`?

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -8,56 +8,54 @@ import kotlin.random.Random
...
+                .map { it.map { it.second } }
                 .toList()
 
-        apps.flatten().forEachIndexed { index, application ->
-            val filteredPeers = peers.map { peerset -> peerset.filter { it != "localhost:${application.getBoundPort()}" } }
-            application.setOtherPeers(filteredPeers)
+        apps.flatten().zip(peers.flatten()).forEachIndexed { index, (app, peer) ->
+            if (index + 1 !in appsToExclude) {
+                app.setOtherPeers(
+                    peers.map { it.filterNot { it == peer } }
```
Rados13, line 56: I think that removing peer peerset or peersets is very common, so we could make it a function.

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
```diff
@@ -157,37 +129,39 @@ class SinglePeersetIntegrationTest {
                 }
                 println("Other peer sent response to ft-agree: $response")
```
Rados13, line 130: Is this needed?

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
```diff
@@ -231,37 +201,37 @@ class SinglePeersetIntegrationTest {
...
+        )
 
-        // application will start
-        delay(5000)
+        val apps = TestApplicationSet(1, listOf(5),
+            actions = mapOf(
+                1 to firstLeaderCallbacks,
+                3 to peer3Callbacks
+            ),
+            configOverrides = (1..5).zip(List(5) { configOverrides }).toMap()
```
Rados13, line 219: What do you think about moving this to a variable like `peerIdToNewConfig`?

Rados13 (changes requested):

src/test/kotlin/com/github/davenury/ucac/consensus/ConsensusSpec.kt
```diff
@@ -89,4 +95,7 @@ class ConsensusSpec {
         }.let { objectMapper.readValue<HistoryDto>(it) }
             .changes.map { ChangeWithAcceptNum(it.change.toChange(), it.acceptNum) }
 
+    private val peer1 = "http://localhost:8081"
+    private val peer2 = "http://localhost:8082"
```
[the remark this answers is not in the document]
Davenury: It's consensus, I didn't deal with it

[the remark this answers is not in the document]
Rados13: Ok so it is a result of wrong rebase

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
Rados13: What about this?

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
Rados13: Ok

Rados13 (changes requested):
Also, tests fail and there are still some conflicts.

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
Davenury: yeah, I don't think so. For now (as far as I could see) we've used this mechanizm twice, every time it was in different context and with slight different approach, as some other outcome was needed - I don't think we should make a function for that for now.

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
```diff
@@ -141,30 +112,26 @@ class SinglePeersetIntegrationTest {
...
-                app.setOtherPeers(listOf(peers - peers[index]))
-            }
+            val peers = apps.getPeers()
 
             // change that will cause leader to fall according to action
             try {
-                executeChange("http://${peers[0]}/create_change")
-                fail("Didn't work")
+                executeChange("http://${peers[0][0]}/create_change")
+                fail("Change passed")
```
Rados13, line 128: Why did you change the comment?

Rados13 (approved):

src/test/kotlin/com/github/davenury/ucac/api/SinglePeersetIntegrationTest.kt
Davenury: It works better, as "Didn't work" is double negation and here change passes, even though it shouldn't

Rados13: Yeah but now we have fail with the name `Change passed`.

PR #32, opened 2022-08-09
Clean up tests and remove further delays

kjarosh:


Rados13 (dismissed):

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -437,47 +434,62 @@ class MultiplePeersetSpec {
             throw RuntimeException()
         }
 
+        val phaser = Phaser(7)
+        phaser.register()
+
+        val peerApplyCommitted = SignalListener {
+            println("Arrived: ${it.subject.getPeerName()}")
```
Davenury, line 441: Maybe it's time to use logger in tests too?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -437,47 +434,62 @@ class MultiplePeersetSpec {
...
             arrayOf("3", "1"),
+            signalListeners = mapOf(Signal.OnHandlingApplyCommitted to peerApplyCommitted),
             configOverrides = configOverrides,
             mode = TestApplicationMode(3, 1)
         )
 
         // mock of not responding peerset2 - is in config, so transaction leader should wait on their responses
         val app4 = createApplication(
             arrayOf("1", "2"),
+            signalListeners = mapOf(Signal.OnHandlingApplyCommitted to peerApplyCommitted),
```
Davenury, line 468: since we're using this map multiple times maybe it would be better to separate it out to a variable?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -497,15 +509,14 @@ class MultiplePeersetSpec {
...
-        delay(5000)
-
         // when - executing transaction something should go wrong after ft-agree
         expectThrows<ServerResponseException> {
             executeChange("http://${peers[0][0]}/create_change")
         }
 
-        // application should elect recovery leader to perform transaction to the end
-        delay(20000)
+        withContext(Dispatchers.IO) {
```
Davenury, line 517: why do you use `withContext` here?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -571,54 +584,63 @@ class MultiplePeersetSpec {
                 throw RuntimeException()
             }
 
+            val arrivalCount = AtomicInteger(0)
+            val peerApplyEnd = SignalListener {
+                arrivalCount.incrementAndGet()
```
Davenury, line 589: Isn't there just `increment()` method? Since we're not using this variable here, get seems unneccessary.

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -627,33 +649,34 @@ class MultiplePeersetSpec {
             val peers =
                 apps.asSequence()
                     .withIndex()
-                    .groupBy{ it.index < 3 }
+                    .groupBy { it.index < 3 }
                     .values
+                    .asSequence()
```
Davenury, line 654: second time `asSequence()`?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -627,33 +649,34 @@ class MultiplePeersetSpec {
...
                 executeChange("http://${peers[0][0]}/create_change")
             }
 
-            // application should elect recovery leader to perform transaction to the end
-            delay(20000)
+            withContext(Dispatchers.IO) {
+                phaser.awaitAdvanceInterruptibly(phaser.arrive(), 60, TimeUnit.SECONDS)
+            }
+
+            expectThat(arrivalCount.get() + 1).isEqualTo(phaser.registeredParties)
```
Davenury, line 673: why this expect? phaser and `arrivalCount` are incremented at the same time - in signal action.

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -571,54 +584,63 @@ class MultiplePeersetSpec {
...
                 arrayOf("1", "2"),
-                emptyMap(),
+                mapOf(Signal.OnHandlingApplyCommitted to peerApplyEnd),
                 configOverrides = configOverrides,
                 mode = TestApplicationMode(1, 2)
             )
             val app5 = createApplication(
                 arrayOf("2", "2"),
-                emptyMap(),
+                mapOf(Signal.OnHandlingApplyCommitted to peerApplyEnd),
```
Davenury, line 625: here also - maybe a separate variable?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: Nope, there's no such method

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: yup, now it's outdated, removed

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: done

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: done

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: done

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: Because without it, it would be a blocking operation in a non-blocking context, which would lead to e.g. thread starvation

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: I was told a Sequence has a better performance than transforming Collections over and over

build.gradle.kts
```diff
@@ -25,6 +25,8 @@ repositories {
 }
 
 dependencies {
+    implementation("org.slf4j:slf4j-api:1.7.36")
```
Davenury, line 28: It's second logger we're using. Do we really need it?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -26,9 +32,16 @@ import strikt.assertions.isEqualTo
...
 import java.util.*
+import java.util.concurrent.Phaser
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicInteger
 import kotlin.random.Random
 
+@Suppress("HttpUrlsUsage")
 class MultiplePeersetSpec {
+    companion object {
+        private val log = LoggerFactory.getLogger(MultiplePeersetSpec::class.java)!!
```
Davenury, line 43: Can this logger be nullable?

Davenury: Also, everywhere in code we're using `logger` name variable. It's clearly nothing, but we could stick to that convention.

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -437,75 +442,97 @@ class MultiplePeersetSpec {
...
+
+        val peerApplyCommitted = SignalListener {
+            log.info("Arrived: ${it.subject.getPeerName()}")
+            phaser.arrive()
+        }
+
+        val signalListenersForLeaders = mapOf(
+            Signal.BeforeSendingApply to failAction,
+        )
+        val signalListenersForNonLeaders = mapOf(
```
Davenury, line 456: Two little suggestions (you don't have to apply them):
* nonLeader == Cohort
* names of those maps aren't really true. Peer 1 is leader up to the point when he fails and then some other peer will become leader. As for now I don't have any better names for them, so they can stay like this

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
Davenury: I still don't get it. `withContext` works like `async() {}.await()`, [according to this](https://www.baeldung.com/kotlin/withcontext-vs-async-await#comparison). So it's still blocking for the test. Then why do we use it and how is it different from just calling phaser function in `runBlocking` scope?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -564,61 +593,75 @@ class MultiplePeersetSpec {
...
+                phaser.arrive()
+            }
+
+            val signalListenersForLeader = mapOf(
+                Signal.BeforeSendingApply to leaderAction,
+                Signal.OnHandlingApplyCommitted to peerApplyEnd,
+            )
+            val signalListenersForNonLeaders = mapOf(
+                Signal.OnHandlingApplyCommitted to peerApplyEnd,
+            )
```
Davenury, line 615: You could do something like that:
```kotlin
val signalListenerForApply = mapOf(
    Signal.OnHandlingApplyCommitted to peerApplyEnd,
)
val signalForLeader = mapOf(Signal.BeforeSendingApply to leaderAction) + signalListenerForApply
```

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: Well, this is Java code, so Kotlin has no idea whether it can be null or not, hence the `!!`. But I'll stick to the convention

build.gradle.kts
kjarosh: Where is the first one?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: done

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: done

build.gradle.kts
Davenury: 6 lines below
`implementation("ch.qos.logback:logback-classic:$logback_version")`

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
Davenury: Actually, can this even be null (I mean is there a situation where that function in java will return null)?

Davenury (approved):
Delete slf4j dependency and answer my out of curiosity question (😄) and it's good for merge for me.

build.gradle.kts
kjarosh: This is actually a logger implementation, whereas `slf4j-api` is an API. Throughout the application we are using `slf4j-api` (hence `org.slf4j.LoggerFactory` imports), but the only one thing `slf4j-api` does is provide a consistent interface between various logger implementations (e.g. log4j, logback, slf4j-simple).

So, this is the same logger, but if we are using slf4j interface explicitly, `slf4j-api` should be a direct dependency instead of a transitive one. The rule is, you use `slf4j-api` in packages you need to use a logger in, and you use an implementation when you create an application no one depends on --- here, we are doing both.

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
kjarosh: As per the documentation it should not return `null`:

> Return an appropriate Logger instance as specified by the name parameter.

Davenury:
I'm on my phone right now, so consider my review to be "approve" (I don't know how to do it in application 😅)

Rados13 (approved):

PR #33, opened 2022-08-16
peers in change

Davenury:


Davenury:
Right now, leader of transaction is also asking himself. Do we need to exclude him (for now there's no easy way of doing this, since his port is known only after application has started).

src/main/kotlin/com/github/davenury/ucac/Application.kt
```diff
@@ -183,8 +183,11 @@ class Application constructor(
         }
     }
 
-    fun setOtherPeers(otherPeers: List<List<String>>) {
-        gpacProtocol.setOtherPeers(otherPeers)
+    fun setGPACPeers(peers: List<List<String>>) {
+        gpacProtocol.setPeers(peers)
+    }
+
+    fun setConsensusOtherPeers(otherPeers: List<List<String>>) {
```
Rados13, line 190: IMO by passing `otherPeers` as `List<List<String>>` gives us a conviction that `otherPeers` in Consensus are stored `List<List<String>>`, which is not true.

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
```diff
@@ -54,42 +55,47 @@ data class ChangeDto(val properties: Map<String, String>) {
 @JsonIgnoreProperties(ignoreUnknown = true)
 sealed class Change(val operation: Operation) {
     abstract fun toDto(): ChangeDto
+    abstract val peers: List<List<String>>
```
Rados13, line 58: Did you decide to add peers in Change?

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -117,7 +117,7 @@ class GPACProtocolImpl(
             logger.info("${getPeerName()} - my state: ${historyManagement.getState()}")
 
             if (message.acceptVal == Accept.COMMIT) {
-                signal(Signal.OnHandlingApplyCommitted, transaction)
+                signal(Signal.OnHandlingApplyCommitted, transaction, message.change.peers)
```
Rados13, line 120: Why didn't you use here `getPeersFromChange` method?

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -60,16 +60,25 @@ class TestApplicationSet(
...
         apps.flatten().forEach { it.stop(gracePeriodMillis, timeoutPeriodMillis) }
     }
 
+    fun setAppsPeers(peers: List<List<String>>) {
+        this.apps
+            .flatten()
+            .filterIndexed { index, _ -> !appsToExclude.contains(index + 1) }
+            .forEach { it.setGPACPeers(peers) }
+    }
+
```
Rados13, line 81: I think that we should here set also consensus peers.

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
Davenury: It's subset of peers but I decided to leave it as nested lists in order to know which peer belongs to which peerset. As I think of it now, `Map<Int, String>` would be better solution, where key is peerset number

src/main/kotlin/com/github/davenury/ucac/Application.kt
```diff
@@ -183,8 +183,11 @@ class Application constructor(
...
         consensusProtocol.setOtherPeers(otherPeers[mode.peersetId - 1])
     }
```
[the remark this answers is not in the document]
Davenury: Consensus Peers are the ones from your peerset (at least that's my understanding) so we don't want to pass all peers. On the other hand, GPAC other peers are all peers in all peersets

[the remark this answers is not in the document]
Rados13: Yes, but with two functions you can setGPAC peers and don't change consensusPeers, which will lead to this that GPAC can see different peerset than consensus.

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
Rados13: Ok, a map will look better or you can use a list of pairs of change and peerset id.

src/main/kotlin/com/github/davenury/ucac/Application.kt
Davenury: Literally one comment ago you said otherwise 😄

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
Davenury: I decided to leave it as list of strings. Since we're using the same dto for internal and external (user) communication, we cannot enforce user to know peersetId - we are open though to creating changes with more than two peersets in that version (even though it's no use now 😄)

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -152,26 +154,28 @@ class GPACProtocolImpl(
     override suspend fun performProtocolAsLeader(
         change: ChangeDto
     ) {
-        val electResponses = electMePhase(change, { responses -> superSet(responses) })
+        logger.info("Peer ${getPeerName()} starts performing GPAC")
+        val enrichedChange = change.copy(peers = change.peers.toMutableList().also { it.add(myAddress) })
```
Davenury, line 158: I think that it could be guarded - if someone passes leader address to `peers` it would still enrich itself and list would look like:
`lisfOf(leader, leader)` which will transform to:
`listOf(peersOfPeerset1, peersOfPeerset1)` -> not quite breaking change but still will cause additional network calls (which may fail BTW).

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -289,51 +295,59 @@ class GPACProtocolImpl(
...
+        val allShards = peers.size >= responses.size / divider.toDouble()
 
         return responses.withIndex()
             .all { (index, value) ->
-                val positiveResponses = if (index + 1 == myPeersetId) value.size + 1 else value.size
                 val allPeers =
-                    if (index + 1 == myPeersetId) peersInTransaction[index].size + 1 else peersInTransaction[index].size
-                positiveResponses > allPeers / 2F
+                    if (index + 1 == myPeersetId) peers[index].size - 1 else peers[index].size
+                value.size >= allPeers / 2F
```
Davenury, line 329: Is this formula correct? As far as I remember it was changed due to having leader in `otherPeers` but now he isn't there anymore.

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -289,51 +295,59 @@ class GPACProtocolImpl(
...
             } && allShards
     }
 
-    companion object {
-        private val logger = LoggerFactory.getLogger(GPACProtocolImpl::class.java)
+    private fun getPeersFromChange(change: ChangeDto): List<List<String>> {
+        return change.peers.map { peer ->
+            if (peer == myAddress) return@map allPeers[myPeersetId]!!
```
Davenury, line 335: `allPeers[myPeersetId]` doesn't include leader, right?

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -289,51 +295,59 @@ class GPACProtocolImpl(
...
+            allPeers.values.find { it.contains(peer) }
+                ?: throw java.lang.IllegalArgumentException("Peer $peer is not found in any of peersets")
```
Davenury, line 337: We should make domain exception for that - we can then return 4xx instead of 5xx, since it's clearly user / higher instance microservice problem (discovery of peers doesn't work).

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -131,12 +131,14 @@ class MultiplePeersetSpec {
         val appsToExclude = listOf(3, 6, 7, 8)
         val apps = TestApplicationSet(2, listOf(3, 5), appsToExclude = appsToExclude)
         val peers = apps.getPeers()
+        val otherPeer = apps.getPeers()[1][0]
 
         delay(5000)
```
Davenury, line 136: What is it doing here?

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
```diff
@@ -253,27 +257,26 @@ class MultiplePeersetSpec {
...
         }
 
         withContext(Dispatchers.IO) {
             phaser.awaitAdvanceInterruptibly(phaser.arrive(), 60, TimeUnit.SECONDS)
         }
 
-        peers.flatten().forEach {
-            askForChanges("http://$it")
+        (0..1).forEach {
+            peers[it].map { askForChanges("http://$it") }.firstOrNull { it.isNotEmpty() }
```
Davenury, line 272: Why is that so?

src/test/kotlin/com/github/davenury/ucac/domain/OperationSpec.kt
```diff
@@ -56,17 +56,17 @@ class OperationSpec {
 
     @Test
     fun `should throw missing parameter exception, when that's the case`() {
-        val changeDto = ChangeDto(mapOf("operation" to "ADD_RELATION"))
+        val changeDto = ChangeDto(mapOf("operation" to "ADD_RELATION"), listOf())
 
-        expectThrows<MissingParameterException> {
+        expectThrows<java.lang.NullPointerException> {
```
Davenury, line 61: I think we should include domain exception here, since `NullPointerException` won't tell us anything in stress tests if it ever comes to it (throwing exception in stress tests)

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
Davenury: Yes, in creating `GPACProtocolImpl` (`Application.kt`, line 96) we filter out our address based on his position in config.

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
Davenury: As I wrote in comment somewhere, there's no point in waiting for consensus to propagate this change - this test doesn't concern consensus mechanizm and should not test (or be dependent) on it. It is required for at least one peer to actually know about this change to propagate it further and that's what this test checks.

src/test/kotlin/com/github/davenury/ucac/api/MultiplePeersetSpec.kt
Davenury: It's delay in disabled test, so for now I guess it's ok for it to be here.

kjarosh (approved):

PR #36, opened 2022-10-03
Rewrite history management

kjarosh:


src/main/kotlin/com/github/davenury/ucac/history/History.kt
```diff
@@ -0,0 +1,62 @@
...
+    private val initialEntry: InitialHistoryEntry = InitialHistoryEntry
+    private val currentEntryId: AtomicReference<String> = AtomicReference(initialEntry.getId())
+    private val entries: ConcurrentHashMap<String, HistoryEntry> = ConcurrentHashMap()
+
+    init {
+        entries[initialEntry.getId()] = initialEntry
+    }
+
+    fun getCurrentEntry(): HistoryEntry {
+        return entries[currentEntryId.get()]!!
```
Davenury, line 19: As much as it's correct, I'd use something like
```kotlin
entries[currentEntryId.get()] ?: throw <our custom exception>
```
it will be better for debug purposes

src/main/kotlin/com/github/davenury/ucac/history/HistoryEntry.kt
```diff
@@ -0,0 +1,9 @@
+package com.github.davenury.ucac.history
+
+interface HistoryEntry {
+    fun getId(): String
+
+    fun getParentId(): String?
+
+    fun serialize(): String
```
Davenury, line 8: What is the puropse of this method?

Davenury (approved):

src/main/kotlin/com/github/davenury/ucac/history/HistoryEntry.kt
```diff
@@ -0,0 +1,30 @@
...
+        fun deserialize(serialized: String): HistoryEntry {
+            val map = ObjectMapper().readValue(serialized, HashMap::class.java)
+            val entry =
+                if (map.containsKey("parentId")) {
+                    IntermediateHistoryEntry(map["content"] as String, map["parentId"] as String)
+                } else {
+                    InitialHistoryEntry
+                }
+            if (entry.serialize() != serialized) {
+                throw IllegalArgumentException("Invalid data: $serialized")
```
Davenury, line 25: Maybe it will be overdone but I'd go step further - create custom exception that we could handle in Application.kt with `exception<>` function (we'll be able to set different http responses etc. unless we will catch everything in higher layers of code)

Davenury (approved):

PR #39, opened 2022-10-13
Use History based on ledger

kjarosh:


src/main/kotlin/com/github/davenury/ucac/common/commonRouting.kt
```diff
@@ -28,11 +30,15 @@ fun Application.commonRouting(
...
+            call.respond(Changes.fromHistory(consensusProtocol.getState()))
         }
 
         get("/consensus/change") {
-            call.respond(consensusProtocol.getState()?.lastOrNull() ?: HttpStatusCode.BadRequest)
+            call.respond(consensusProtocol.getState()
+                .getCurrentEntry()
+                .takeIf { it != InitialHistoryEntry }
+                ?.let { Change.fromHistoryEntry(it) }
+                ?: HttpStatusCode.BadRequest)
```
Davenury, line 41: Bad request? Imo not found would be better

src/main/kotlin/com/github/davenury/ucac/common/commonRouting.kt
```diff
@@ -28,11 +30,15 @@ fun Application.commonRouting(
...
-            call.respond(Changes(consensusProtocol.getState() ?: listOf()))
+            call.respond(Changes.fromHistory(consensusProtocol.getState()))
         }
 
         get("/consensus/change") {
-            call.respond(consensusProtocol.getState()?.lastOrNull() ?: HttpStatusCode.BadRequest)
+            call.respond(consensusProtocol.getState()
+                .getCurrentEntry()
+                .takeIf { it != InitialHistoryEntry }
+                ?.let { Change.fromHistoryEntry(it) }
```
Davenury, line 40: Don't you have a function for that? `getLastChange()` I think

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/HistoryStateMachine.kt
```diff
@@ -31,7 +31,9 @@ class HistoryStateMachine(override var state: History = mutableListOf()) :
     enum class OperationType {
         LAST {
             override fun invokeOperation(state: History): String {
-                return state.last().let { objectMapper.writeValueAsString(it) }
+                return state.getCurrentEntry()
+                    .let { Change.fromHistoryEntry(it) }
```
Davenury, line 35: Here we aren't checking if the history entry is the initial history entry. Why?

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
```diff
@@ -3,20 +3,36 @@ package com.github.davenury.ucac.common
...
 // see https://github.com/FasterXML/jackson-databind/issues/2742#issuecomment-637708397
 class Changes : ArrayList<Change> {
     constructor() : super()
+
     constructor(collection: List<Change>) : super(collection)
+
+    fun toHistory(): History {
+        val h = History()
+        this.map { it.toHistoryEntry() }
+            .forEach { h.addEntry(it) }
```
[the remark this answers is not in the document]
kjarosh: Because `History` is needed and not `List<HistoryEntry>`

src/main/kotlin/com/github/davenury/ucac/common/commonRouting.kt
kjarosh: Sure, I just left the status as it was before

src/main/kotlin/com/github/davenury/ucac/common/commonRouting.kt
kjarosh: Sure, this function is within `HistoryManagement`. I think using it would mean a bigger refactor (which I am planning to do in next PRs)

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/HistoryStateMachine.kt
kjarosh: I just made sure to retain the behavior which was before. In this case an exception will be thrown (as in the previous implementation).

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
```diff
@@ -3,20 +3,36 @@ package com.github.davenury.ucac.common
...
+        return h
+    }
```
Rados13, line 26: What do you think about change like this?
```suggestion
    fun toHistory(): History = this.fold(History()){h, change -> h.addEntry(change.toHistoryEntry()) }
```

src/main/kotlin/com/github/davenury/ucac/common/InMemoryHistoryManagement.kt
```diff
@@ -26,15 +28,15 @@ class InMemoryHistoryManagement(
...
-            consensusProtocol.getState()?.last()
-        } catch (ex: java.util.NoSuchElementException) {
-            logger.error("History is empty!")
-            null
+    override fun getLastChange(): Change? {
+        val currentEntry = consensusProtocol.getState().getCurrentEntry()
+        if (currentEntry == InitialHistoryEntry) {
+            return null
         }
+        return currentEntry.let { Change.fromHistoryEntry(it) }
```
Rados13, line 36:
```suggestion
        return if (currentEntry == InitialHistoryEntry) {
            null
        } else {
        currentEntry.let { Change.fromHistoryEntry(it) }
        }
```

src/main/kotlin/com/github/davenury/ucac/consensus/raft/domain/Ledger.kt
```diff
@@ -34,25 +34,30 @@ data class Ledger(
...
         val newId = previousId + 1
-        proposedItems.add(LedgerItem(newId,iteration, change))
+        proposedItems.add(LedgerItem(newId, iteration, change))
         return newId
     }
 
-    fun getHistory(): History =
-        (acceptedItems + proposedItems).map { it.change }.toMutableList()
+    fun getHistory(): History {
+        // TODO why proposed??
```
Rados13, line 42: If only I could remember.

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
kjarosh: My attitude towards this change is ambivalent. However, it does not compile

src/main/kotlin/com/github/davenury/ucac/common/InMemoryHistoryManagement.kt
kjarosh: Made it even more functional

src/main/kotlin/com/github/davenury/ucac/common/Change.kt
Rados13: Ok, so add TODO that it should be changed.

Rados13 (approved):

PR #41, opened 2022-10-15
docs and command fixes

Davenury:
* script that installs kubectl, minikube, golang, helm, if not installed locally.
* same script adds helm chart repositories
* updated `k8s-stuff` README
* command REAMDE
* some command fixes (e.g. flag `--create-namespace` tried to create the same namespace couple of times, change context)


kjarosh (approved):

Rados13 (changes requested):
Ok I was during the review of this PR but as you don't care about my opinion I won't finish it :cry:

PR #43, opened 2022-10-19
Clean up config

kjarosh:


src/main/kotlin/com/github/davenury/ucac/ApplicationMode.kt
```diff
@@ -166,40 +112,21 @@ object KubernetesApplicationMode: ApplicationMode() {
...
-            throw IllegalStateException()
-        }
 }
 
+private val log = LoggerFactory.getLogger(ApplicationMode::class.java)
+
 fun determineApplicationMode(args: Array<String>): ApplicationMode {
-    val applicationMode = when (System.getenv("application_environment")) {
+    val applicationModeString = System.getenv("application_environment")
+    log.info("Using application mode $applicationModeString")
```
Rados13, line 123: Do we need this log on the info level?

src/main/kotlin/com/github/davenury/ucac/config.kt
```diff
@@ -1,25 +1,37 @@
...
 import com.sksamuel.hoplite.addResourceSource
 import java.time.Duration
 
-data class Config(val raft: RaftConfig, val peers: PeersConfig, val protocol: ProtocolConfig, val applicationEnv: String?)
-data class PeersConfig(val peersAddresses: List<List<String>>, val maxLeaderElectionTries: Int)
+data class Config(
+    val host: String = "",
+    val port: Int = -1,
+    val peerId: Int = -1,
+    val peersetId: Int = -1,
```
Rados13, line 12: Are we sure about this `-1`/`""`. Wouldn't it be better to use Optional, as Kotlin has great support for it?

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -25,11 +25,12 @@ class RaftConsensusProtocolImpl(
     private val peersetId: Int,
     private var peerAddress: String,
     private val ctx: ExecutorCoroutineDispatcher,
-    private var otherConsensusPeers: List<String>,
+    private val allPeers: List<String>,
```
Rados13, line 28: Are you sure about this field? After calling `setOtherPeers` this field can become totally outdated.

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/Constants.kt
```diff
@@ -25,38 +24,28 @@ import java.util.*
  * limitations under the License.
  */
 
-class RaftConfiguration(
+class RatisConfiguration(
     peersetId: Int,
-    overrides: Map<String, Any> = emptyMap()
+    ratisConfig: RatisConfig,
 ) {
```
Rados13, line 30: Are you sure about this naming? Because for me it's a bit strange that the class `RatisConfiguration` is use in the constructor variable of type `RatisConfig`.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -25,11 +25,12 @@ class RaftConsensusProtocolImpl(
...
     private val signalPublisher: SignalPublisher = SignalPublisher(emptyMap()),
     private val protocolClient: RaftProtocolClient,
     private val heartbeatTimeout: Duration = Duration.ofSeconds(4),
-    private val heartbeatDelay: Duration = Duration.ofMillis(500)
+    private val heartbeatDelay: Duration = Duration.ofMillis(500),
+    private var otherConsensusPeers: List<String> = allPeers.filter { it != peerAddress },
```
Davenury, line 33: can you actually do that? to calculate other arg from another argument of the same function. Looks pretty sus to me (and not readable at all). Imo - better calculate it inside (in init or something)

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/Constants.kt
```diff
@@ -25,38 +24,28 @@ import java.util.*
...
-        configPeers = if (config.applicationEnv == "k8s") {
-            System.getenv()["RATIS_PEERS_ADDRESSES"].let { objectMapper.readValue(it, ArrayList<ArrayList<String>>()::class.java) }
-        } else {
-            config.raft.server.addresses
-        }
-
-        PATH = config.raft.server.root.storage.path
-        CLUSTER_GROUP_ID = UUID.fromString(config.raft.clusterGroupIds[peersetId - 1])
-        PEERS = configPeers[peersetId - 1].mapIndexed { index, address ->
+        clusterGroupId = UUID(0, peersetId.toLong())
```
Davenury, line 39: I don't think it'll work or at least it will destroy idea of two separate ratis instances. That's how ratis actually determines, if two peers are in the same group. Since we allow it, I think we are able to delete ratis.

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/Constants.kt
```diff
@@ -25,38 +24,28 @@ import java.util.*
...
+        peers = ratisConfig.addresses[peersetIndex].mapIndexed { index, address ->
```
Davenury, line 40: You did use it on k8s, right? Becuase I have a strange deja vu, since library we're using is unable to parse list of lists (which we exactly face here) from env variables, so it won't work on k8s. But maybe I'm wrong.

src/main/kotlin/com/github/davenury/ucac/Application.kt
```diff
@@ -44,60 +43,73 @@ fun createApplication(
     configOverrides: Map<String, Any> = emptyMap(),
     mode: ApplicationMode = determineApplicationMode(args)
 ): Application {
-    return Application(signalListeners, configOverrides, mode)
+    var config = loadConfig(configOverrides)
+
+    // TODO temporary, for backward compatibility
```
Davenury, line 48: will this TODO be resolved in this PR?

src/main/kotlin/com/github/davenury/ucac/ApplicationMode.kt
```diff
@@ -1,42 +1,35 @@
...
-    // should be change in application
-    override val otherPeers: List<List<String>>
-        get() = listOf(listOf())
-    override val logger: Logger = LoggerFactory.getLogger(TestApplicationMode::class.java)
 
     override fun name(): String = "TestApplicationMode"
 }
 
 object DockerComposeApplicationMode: ApplicationMode() {
+    private val log = LoggerFactory.getLogger(DockerComposeApplicationMode::class.java)
```
Davenury, line 27: It's nothing, but we're using `logger` everywhere, so for the sake of convention, please change it

src/main/kotlin/com/github/davenury/ucac/ApplicationMode.kt
```diff
@@ -166,40 +112,21 @@ object KubernetesApplicationMode: ApplicationMode() {
         host = me
         port = 8080
         peersetId = _peersetId
-        otherPeers = getOtherPeers(config, me)
-        nodeId = id
+        peerId = id
     }
-
-    private fun getOtherPeers(config: Config, me: String): List<List<String>> =
```
Davenury, line 173: Question: I don't see anywhere, where do you actually calculate `otherPeers` - don't we need them anymore?

src/main/kotlin/com/github/davenury/ucac/ApplicationMode.kt
```diff
@@ -166,40 +112,21 @@ object KubernetesApplicationMode: ApplicationMode() {
...
-            }
-        } catch (e: java.lang.IndexOutOfBoundsException) {
-            logger.error(
-                "Peers addresses doesn't have enough elements in list - peers addresses length: ${config.peers.peersAddresses.size}, index: ${peersetId - 1}"
-            )
-            throw IllegalStateException()
-        }
 }
 
+private val log = LoggerFactory.getLogger(ApplicationMode::class.java)
```
Davenury, line 119: Since you're creating logger here, couldn't you use it in `*ApplicationMode`?

src/main/kotlin/com/github/davenury/ucac/config.kt
Davenury: +1 - I'm insecure about using -1 and empty strings (maybe PTSD from operative systems, but still) - IMO you could use nullables (I think it will even simplify these ifs in `Application.kt`)

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
```diff
@@ -26,7 +26,7 @@ class TestApplicationSet(
         )
 
         val addressesOverrides = mapOf(
-            "peers.peersAddresses" to (1..numberOfPeersets).map { (1..numberOfPeersInPeersets[it - 1]).map { nonRunningPeer } }
+            "peerAddresses" to (1..numberOfPeersets).map { (1..numberOfPeersInPeersets[it - 1]).map { nonRunningPeer } }
```
Davenury, line 29: I think those values up there (`ratisConfigOverrides`) have changed as well

Davenury (changes requested):
I would love to believe that but I need to ask - did you actually run application in all modes (local, compose and k8s) and checked if your changes are working (2 peersets, 3 peers in each, 2 changes sent). I know tests are tests but we needed to have multiple changes because of tests (we're setting peers in tests e.g.).

src/main/kotlin/com/github/davenury/ucac/ApplicationMode.kt
kjarosh: When you configure this app in a wrong way (e.g. by supplying wrong application mode), you will get a weird error message. So, until we change how the application is configured I would prefer currently to leave it at `info`.

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/Constants.kt
kjarosh: Removed `Constants.kt`

src/main/kotlin/com/github/davenury/ucac/Application.kt
kjarosh: It will be resolved in the next one

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/Constants.kt
kjarosh: How will it destroy the idea of two separate instances?

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
kjarosh: Should have been a parameter

kjarosh: Actually, I decided to make `otherConsensusPeers` a function

src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt
kjarosh: Fixed as well

src/main/kotlin/com/github/davenury/ucac/ApplicationMode.kt
kjarosh: The idea of the simplification of the configuration is to configure peers instead of calculating them, k8s will require a major configuration overhaul

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
kjarosh: This fragment has been rewritten

.github/workflows/go.yml
```diff
@@ -0,0 +1,18 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '>=1.19'
+      - name: Go get
+        working-directory: cmd
+        run: go get
+      - name: Go build
+        working-directory: cmd
+        run: go build -o ucac
```
Davenury, line 18: Delete it, I'm doing much more complex workflow now (including building command, setting up k8s and actually deploying and testing, if change is created)

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/Constants.kt
Davenury: considering that ratis identifies groups by those uuids, I don't think something good will result from 2 groups having the same uuid, but Radek is much more an expert than me in ratis business

Rados13: If I understand everything correctly all nodes from the same peer set have to have the same clusterGroupId and this changes only that it will be generated based on peersetId instead of reading from config. So IMO it shouldn't break anything.

src/main/kotlin/com/github/davenury/ucac/Application.kt
```diff
@@ -30,74 +29,94 @@ import io.netty.channel.socket.nio.NioServerSocketChannel
...
+    val application = createApplication(configOverrides = configOverrides)
+    application.startNonblocking()
+
+    while (!Thread.interrupted()) {
+        try {
+            Thread.sleep(1000)
+        } catch (e: InterruptedException) {
+            break;
+        }
+    }
```
Rados13, line 57: Why do we need this?

kjarosh: It allows to stop the server using an interrupt, which resulted from the fact that I removed `startBlocking` due to compatibility issues with `startNonblocking` (it was not calling `setPeerAddress`).

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -49,20 +48,24 @@ class RaftConsensusProtocolImpl(
...
     private var executorService: ExecutorCoroutineDispatcher? = null
 
+    private fun otherConsensusPeers(): List<String> {
+        return allPeers.filter { it != peerAddress }
+    }
+
     override suspend fun begin() {
-        assert(!otherConsensusPeers.contains(peerAddress))
-        logger.info("$peerId - Start raft on address $peerAddress, other peers: $otherConsensusPeers")
+        assert(!otherConsensusPeers().contains(peerAddress))
```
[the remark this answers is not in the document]
kjarosh: removed

Rados13 (approved):

Davenury (approved):

PR #44, opened 2022-10-19
Fix max retries tests

Rados13:


src/main/kotlin/com/github/davenury/ucac/common/ProtocolTimer.kt
```diff
@@ -22,17 +23,21 @@ class ProtocolTimerImpl(
...
+                val exponent = 1.5.pow(iteration)
+
                 val backoff = (
                         if (backoffBound.isZero) 0
-                        else randomGenerator.nextLong().absoluteValue % backoffBound.toMillis()
+                        else randomGenerator.nextLong().absoluteValue % (backoffBound.toMillis() * exponent).toLong()
                         )
                     .let { Duration.ofMillis(it) }
                 val timeout = delay.plus(backoff)
+                if (iteration != 0) println("Sleeping for: $timeout /-/ $backoff /-/ $exponent /-/ $iteration")
```
kjarosh, line 40: `log.info` + describe the printed parameters

Davenury, line 40: logger

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -152,16 +155,34 @@ class GPACProtocolImpl(
...
             }
-        val electResponses = electMePhase(enrichedChange, { responses -> superSet(responses, getPeersFromChange(enrichedChange)) })
+
+        val electMeResult =
+            electMePhase(enrichedChange, { responses -> superSet(responses, getPeersFromChange(enrichedChange)) })
+
+        if (iteration == maxLeaderElectionTries) {
+            logger.error("Transaction failed due to too many retries of becoming a leader.")
+            signal(Signal.ReachedMaxRetries, transaction, getPeersFromChange(change))
+            return
```
Davenury, line 175: shouldn't we throw exception here? Or do we silently fail here?

src/main/kotlin/com/github/davenury/ucac/common/ProtocolTimer.kt
Rados13: I will remove it as it was used only for debugging purposes.

Rados13: I will remove it as it was used only for debugging purposes.

src/main/kotlin/com/github/davenury/ucac/common/commonRouting.kt
```diff
@@ -17,10 +18,20 @@ fun Application.commonRouting(
...
+            call.respond(transactionResult)
+        }
+
         post("/create_change") {
             val change = call.receive<Change>()
-            gpacProtocol.performProtocolAsLeader(change)
-            call.respond(HttpStatusCode.OK)
+            val statusCode = when (gpacProtocol.performProtocolAsLeader(change)) {
+                TransactionResult.DONE -> HttpStatusCode.OK
+                TransactionResult.FAILED -> HttpStatusCode.BadRequest
```
Davenury, line 31: It's nothing but transaction failed isn't the fault of user (I think), so probably it should be 500 or so

Davenury (approved):

PR #48, opened 2022-10-23
Remove proposed items from history

kjarosh:


src/main/kotlin/com/github/davenury/ucac/consensus/HistoryManagementRouting.kt
```diff
@@ -12,7 +13,11 @@ fun Application.historyManagementRouting(historyManagement: HistoryManagement) {
         route("/change") {
             get {
                 val result = historyManagement.getLastChange()
-                call.respond(result ?: ErrorMessage("Error any change doesn't exist"))
+                if (result != null) {
+                    call.respond(result)
+                } else {
+                    call.respond(HttpStatusCode.NotFound, ErrorMessage("Error any change doesn't exist"))
+                }
```
Davenury, line 20: You could do 
```kotlin
result?.let { call.respond(it) }  ?: call.respond(...)
```

Davenury (approved):

Rados13 (approved):

PR #50, opened 2022-10-24
Start peer IDs and peerset IDs from 0

kjarosh:


cmd/commands/deploy.go
```diff
@@ -45,7 +45,7 @@ func CreateDeployCommand() *cobra.Command {
 
 					peerConfig := PeerConfig{
 						PeerId:         peerId,
-						PeersetId:      strconv.Itoa(idx + 1),
+						PeersetId:      strconv.Itoa(idx),
```
Davenury, line 48: It seemes I cannot comment on line that wasn't change, so in line 43, we start iterating from one up to num (included). I think it should be changed too.

kjarosh: Fixed, deployment now works

PR #51, opened 2022-10-24
Separate GPAC and consensus

kjarosh:


Rados13 (commented):
Some tests check only in one peer from peerset if a change is applied. Instead, we should check if all peers from peerset have applied this change. Let's change it in this PR.

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
```diff
@@ -132,7 +132,7 @@ class GPACProtocolImpl(
                 signal(Signal.OnHandlingApplyCommitted, transaction, message.change)
             }
             if (message.acceptVal == Accept.COMMIT && !transactionWasAppliedBefore()) {
-                historyManagement.change(message.change)
+                historyManagement.getState().addEntry(message.change.toHistoryEntry())
```
Davenury, line 135: That's something I wanted to ask - what's better - calling `getState().addEntry()` every time or to actually include method `addEntry` in `historyManagement`?

Davenury (dismissed):

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
kjarosh: I will remove history management (possibly temporarily) to simplify the code, so it does not matter rn

kjarosh:
> Some tests check only in one peer from peerset if a change is applied. Instead, we should check if all peers from peerset have applied this change. Let's change it in this PR.

Let me know if there are other tests which have to be changed

src/main/kotlin/com/github/davenury/ucac/gpac/domain/GPACProtocol.kt
Davenury: I wanted to ask which one is better in generic scenario, not to request changes on that 😅 
Like which one is better in case where we call method on method - should we add another method in the first class (history Management in this case) or leave it as it is (or it depends). Tbh it didn't matter to me anyway but still that's something I wanted to find out 😂

kjarosh: Oh, sure — generally this specific line of code I wrote violates LoD (the Law of Demeter), which causes tighter coupling. Obviously "it depends", because most of principles talk about concepts instead of actual code, and this is also the case in LoD — it focuses more on the concept of the public interface of HistoryManagement — if History is considered an internal state or a completely independent class, LoD applies; if it's considered a public interface of HistoryManagement, I would not apply LoD here.

Rados13 (approved):

PR #58, opened 2022-11-02
Use image from Gradle workflow in k8s

kjarosh:


.github/workflows/k8s.yml
```diff
@@ -1,12 +1,19 @@
...
+      - completed
 
 jobs:
   build:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
+      - name: Fail if dependent workflows failed
+        run: exit 1
+        if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
Davenury, line 16: You sure about that? If conclusion is success you want to fail?

.github/workflows/k8s.yml
```diff
@@ -1,12 +1,19 @@
...
+    types:
+      - completed
 
 jobs:
   build:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
+      - name: Fail if dependent workflows failed
+        run: exit 1
```
Davenury, line 15: Aren't there any more sophisticated ways to do that? `exit 1` seems... ordinary and put of place here. I think I saw something like fail in github actions documentation.

.github/workflows/k8s.yml
```diff
@@ -1,12 +1,19 @@
 name: K8s
 
-on: [push, pull_request]
+on:
+  workflow_run:
+    workflows: [ "Gradle" ]
```
Davenury, line 5: As we're playing around with workflows, I think we should change name of the one that you point to. Maybe something like "build"?

.github/workflows/k8s.yml
```diff
@@ -20,7 +27,7 @@ jobs:
         run: go get && go build -o cmd
       - name: deploy pods
         working-directory: ./cmd
-        run: ./cmd deploy -n test --create-namespace --peers=3,1 --wait-for-readiness
+        run: ./cmd deploy -n test --create-namespace --peers=3,1 --wait-for-readiness --image ghcr.io/davenury/davenury/ucac:${{ github.ref == 'refs/heads/main' && 'latest' || github.sha }}
```
Davenury, line 30: I'm quite sure that I corrected this mistake (double davenury) and new images are pushed with just one davenury

.github/workflows/gradle.yml
```diff
@@ -26,5 +26,5 @@ jobs:
           github-push-secret: ${{ secrets.DOCKER_DEPLOYMENT_TOKEN }}
           docker-image-name: ucac
           docker-image-tag: "${{ github.ref == 'refs/heads/main' && 'latest' || github.sha }}"
-          dockerfile-path: ./Dockerfile # TODO - change after two separate dockerfiles
+          dockerfile-path: ./Dockerfile
```
Davenury, line 29: Is this TODO resolved - I mean did you merge PR that creates two separate dockerfiles?

Davenury (changes requested):

PR #59, opened 2022-11-05
Gpac sync async api and worker

Rados13:


src/main/kotlin/com/github/davenury/ucac/ApplicationUcac.kt
```diff
@@ -60,24 +60,25 @@ fun main(args: Array<String>) {
...
-        val logger = LoggerFactory.getLogger(Application::class.java)
+        val logger = LoggerFactory.getLogger(ApplicationUcac::class.java)
     }
 
     private val engine: NettyApplicationEngine
     private var raftNode: HistoryRatisNode? = null
     private var consensusProtocol: RaftConsensusProtocol? = null
     private val ctx: ExecutorCoroutineDispatcher = Executors.newCachedThreadPool().asCoroutineDispatcher()
     private lateinit var gpacProtocol: GPACProtocol
+    private var service: ApiV2Service? = null
```
Davenury, line 81: You are initializing it in init - wouldn't `private val service: ApiV2Service` suffice?

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Routing.kt
```diff
@@ -13,60 +13,68 @@ import java.time.Duration
...
+                        changeApplied = "unknown",
+                    ),
+                )
+            }
+        }
+    }
+
     routing {
         post("/v2/change/async") {
+            val enforceGpac: Boolean = call.request.queryParameters["enforce_gpac"] != null
```
Davenury, line 64: I'd rather add `...["enforce_gpac"]?.toBoolean() == true` (or sth like that). Generic clients may use it and setting explicitly `true` is IMO better than relying on not being present at all (note that ?enforce_gpac=null will also result in enforcing gpac, so generic function in tests is out of question)

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -29,26 +41,39 @@ class ApiV2Service(
...
-    suspend fun addChange(change: Change): CompletableFuture<ChangeResult> {
-        val peers = change.peers
-        return if (allPeersFromMyPeerset(peers)) {
-            consensusProtocol.proposeChangeAsync(change)
-        } else {
-            gpacProtocol.proposeChangeAsync(change)
-        }
+    fun getChangeStatus(changeId: String): CompletableFuture<ChangeResult> = consensusProtocol.getChangeResult(changeId)
+        ?: gpacProtocol.getChangeResult(changeId)
+        ?: throw RuntimeException("Change doesn't exist")
```
Davenury, line 46: I'd add custom exception and register it in application

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -29,26 +41,39 @@ class ApiV2Service(
...
+
+    suspend fun addChange(change: Change, enforceGpac: Boolean = false): CompletableFuture<ChangeResult> {
+        val consensusChange: Boolean = allPeersFromMyPeerset(change.peers) && !enforceGpac
+        val cf = CompletableFuture<ChangeResult>()
```
Davenury, line 50: You could do it like this:
```kotlin
return CompletableFuture<ChangeResult>()
  .also {
    queue.addLast(ProcessorJob(change, it, !consensusChange))
    channel.send(Unit)
  }
```
In that way, you'd get rid of this "beautiful" variable name `cf`.

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -29,26 +41,39 @@ class ApiV2Service(
...
+        queue.addLast(ProcessorJob(change, cf, !consensusChange))
```
Davenury, line 51: Maybe it's just me, but I'd name the last argument, as right now, I don't know why you pass negation of `consensusChange`

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -29,26 +41,39 @@ class ApiV2Service(
...
-            consensusProtocol.proposeChangeAsync(change)
-        } else {
-            gpacProtocol.proposeChangeAsync(change)
-        }
+    fun getChangeStatus(changeId: String): CompletableFuture<ChangeResult> = consensusProtocol.getChangeResult(changeId)
+        ?: gpacProtocol.getChangeResult(changeId)
+        ?: throw RuntimeException("Change doesn't exist")
+
+    suspend fun addChange(change: Change, enforceGpac: Boolean = false): CompletableFuture<ChangeResult> {
+        val consensusChange: Boolean = allPeersFromMyPeerset(change.peers) && !enforceGpac
```
Davenury, line 49: By convention - `isOnlyConsensusChange` or sth like that (boolean variables should have verb in them that the state of this variable answers)

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -29,26 +41,39 @@ class ApiV2Service(
...
+        withTimeout(timeout ?: config.rest.defaultSyncTimeout) {
+            addChange(change, enforceGpac).await()
         }
+    } catch (e: TimeoutCancellationException) {
+        null
+    }
+
+
+    fun setPeers(peers: Map<Int, List<String>>, myAddress: String) {
+        val myPeerset = peers[config.peersetId]!!.plus(myAddress)
```
Davenury, line 71: Why `peers[config.peersetId]` wouldn't contain `myAddress`?

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ConcurrentLinkedDeque
+import java.util.concurrent.locks.ReentrantLock
+
+
+class Worker(
+    private val queue: Deque<ProcessorJob>,
+    private val channel: Channel<Unit>,
+    private val gpacProtocol: GPACProtocol,
+    private val consensusProtocol: ConsensusProtocol,
```
Davenury, line 20: Didn't we talk about a class that contains all the implementations of consensus and AC protocols?

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -0,0 +1,56 @@
...
+
+
+    private suspend fun processingQueue() {
+        try {
+            while (!Thread.interrupted()) {
+                while (queue.isEmpty()) channel.receive()
+                val job = queue.pop()
+                val result =
+                    if (job.gpacChange) gpacProtocol.proposeChangeAsync(job.change)
+                    else consensusProtocol.proposeChangeAsync(job.change)
```
Davenury, line 31: Now I see, why you used negation a couple of comments earlier. But wouldn't it be more readable, if you used `job.onlyConsensusChange`?

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -0,0 +1,56 @@
...
+                job.completableFuture.complete(result.await())
+            }
+        } catch (e: Exception) {
+            when (e) {
+                is InterruptedException -> {
+                    println("Worker interrupted")
```
Davenury, line 37: logger

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -0,0 +1,56 @@
...
+                    processingQueue()
+                }
+
+                else -> processingQueue()
```
Davenury, line 41: I don't see a reason for `when` here, if no matter what happens, you are calling `processingQueue`.

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun run() = runBlocking {
+        processingQueue()
+    }
+
+}
+
+data class ProcessorJob(
+    val change: Change,
+    val completableFuture: CompletableFuture<ChangeResult>,
+    val gpacChange: Boolean = true
```
Davenury, line 55: As I mentioned earlier, you could use "isConsensusOnly" or something like that.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -1,433 +1,470 @@
-package com.github.davenury.ucac.commitment.gpac
```
Davenury, line 1: Has anything changed in here, or is it just "my encoding is better than yours" thing?

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocolClient.kt
```diff
@@ -62,6 +62,7 @@ class GPACProtocolClientImpl : GPACProtocolClient {
             peersets.map {
                 CoroutineScope(Dispatchers.IO).async {
                     val (httpResult, value) = gpacHttpCall<K, T>("http://$it/$urlPath", requestBody, errorMessage)
+                    println("Acc: $acc, value: $value")
```
Davenury, line 65: Is it necessary? If yes, I'd opt for logger.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -368,28 +368,32 @@ class RaftConsensusProtocolImpl(
...
-        asyncProposeChangeToLedger(change).await()
-
     //  TODO: sync change will have to use Condition/wait/notifyAll
-    private suspend fun asyncProposeChangeToLedger(change: Change): CompletableFuture<ChangeResult> {
+    private suspend fun proposeChangeToLedger(change: Change): CompletableFuture<ChangeResult> {
 
 //      TODO: it will be changed
         val id: Int
 
+        val cf = CompletableFuture<ChangeResult>()
```
Davenury, line 377: Please more descriptive names - why do we need this completable future here?

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -368,28 +368,32 @@ class RaftConsensusProtocolImpl(
...
 
-
-    private suspend fun syncProposeChangeToLedger(change: Change): ChangeResult =
-        asyncProposeChangeToLedger(change).await()
-
     //  TODO: sync change will have to use Condition/wait/notifyAll
-    private suspend fun asyncProposeChangeToLedger(change: Change): CompletableFuture<ChangeResult> {
+    private suspend fun proposeChangeToLedger(change: Change): CompletableFuture<ChangeResult> {
 
 //      TODO: it will be changed
```
Davenury, line 374: to what? and when?

src/test/kotlin/com/github/davenury/ucac/consensus/ConsensusSpec.kt
```diff
@@ -77,15 +78,17 @@ class ConsensusSpec {
         val peerAddresses = peerset.getPeers()[0]
 
         phaser.arriveAndAwaitAdvanceWithTimeout()
+        println("Phaser: ${phaser.phase}")
```
Davenury, line 81: Is this (and in L91) println necessary?

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Routing.kt
```diff
@@ -75,9 +83,15 @@ fun Application.apiV2Routing(
             return@get call.respond(change ?: HttpStatusCode.NotFound)
         }
 
+        get("/v2/change_status/{id}") {
+            val id = call.parameters["id"]!!
+
+            val result = service.getChangeStatus(id).get()
```
kjarosh, line 89: It does not return status of the change — it _waits_ for the change. There's additionally no timeout, so theoretically, it's possible for this endpoint to not return.

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
Rados13: It doesn't contain peer address in tests. I don't know why, because I am not the author of this code, but you can check in git blame who wrote this.
https://github.com/Davenury/Unifying-Consensus-And-Atomic-Commitment-PoC/blob/00574ac48be9fe895ac57890fb2951db5f25d30e/src/test/kotlin/com/github/davenury/ucac/utils/TestApplicationSet.kt#L66-L72
So this is mostly a test fix. But I didn't thought about a situation where `myAddress` is already in it so I will add `distinct` to remove duplicated addresses.

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
Rados13: APIService has to all of the implementations but Worker also needs access to all implementations.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
Rados13: You can filter whitespace in github option
image

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Rados13: I don't remember so I removed it.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Rados13: We need to return from this function CompletableFuture instance, so we first need to initialize one and we did this here in this line.

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
Davenury: Author would be probably me and I don't know what I was smoking when I did this 😁 although wait I know - it's because of we don't want leader of protocol to send requests to himself. And this behavior was (I hope it's still there) in ApplicationUcac.kt, where from config we're excluding "us" (as the peer) from all peers.

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
Davenury: Yeah, that's why I thought that we planned to have a class that would gather all of those implementations (not just service but something on different level, that only responsibility would be to determine which protocol to use, so we wouldn't actually duplicate the logic). Of course it's totally out of the scope of this PR, I thought that it was already implemented.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
Davenury: 😮😮😮 a person learns all their life and (in my case) will forever be stupid. Thanks!

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Davenury: Fair enough but still cf triggers operative systems vibes in me 😂

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Davenury: "Był chłop, ni ma chłopa" 😂😂

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
Rados13: Do you know that you can't exclude an address if it wasn't previously there?

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Routing.kt
```diff
@@ -75,9 +83,18 @@ fun Application.apiV2Routing(
             return@get call.respond(change ?: HttpStatusCode.NotFound)
         }
 
+        get("/v2/change_status/{id}") {
+            val id = call.parameters["id"]!!
+
+            val result = service.getChangeStatus(id)
+
+            if (result.isDone) respondChangeResult(result.get(), call)
+            else call.respond("Change is processed")
```
Davenury, line 92: "change is being processed" like it's still processing, right? (imo you don't have to change it though)

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -6,49 +6,70 @@ import com.github.davenury.ucac.common.ChangeResult
...
 
-    suspend fun addChange(change: Change): CompletableFuture<ChangeResult> {
-        val peers = change.peers
-        return if (allPeersFromMyPeerset(peers)) {
-            consensusProtocol.proposeChangeAsync(change)
-        } else {
-            gpacProtocol.proposeChangeAsync(change)
+    fun getChangeStatus(changeId: String): CompletableFuture<ChangeResult> = consensusProtocol.getChangeResult(changeId)
+        ?: gpacProtocol.getChangeResult(changeId)
+        ?: throw ChangeDoesntExist()
```
Davenury, line 45: I think I'd add at least changeId to this exception but maybe it's overkill.

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -0,0 +1,58 @@
...
+                    if (job.isConsensusOnly) consensusProtocol.proposeChangeAsync(job.change)
+                    else gpacProtocol.proposeChangeAsync(job.change)
+                job.completableFuture.complete(result.await())
+            }
+        } catch (e: Exception) {
+            when (e) {
+                is InterruptedException -> {
+                    logger.debug("Worker interrupted")
+                }
+            }
```
Davenury, line 39: You don't have to change it but you could do it easier by using `if` - `when` is more like `switch` in C and it's overkill here.

src/main/kotlin/com/github/davenury/ucac/consensus/ratis/HistoryRatisNode.kt
```diff
@@ -37,4 +37,8 @@ class HistoryRatisNode(
     override fun getState(): History {
         return history
     }
+
+    override fun getChangeResult(changeId: String): CompletableFuture<ChangeResult> {
+        TODO("Not yet implemented")
+    }
```
Davenury, line 43: ? XD

Davenury (approved):

Davenury (approved):

PR #64, opened 2022-11-08
Two different changes

Davenury:


src/main/kotlin/com/github/davenury/ucac/history/History.kt
```diff
@@ -25,18 +25,24 @@ class History {
...
+    fun addEntry(entry: HistoryEntry, swapParent: Boolean = false) {
         val newId = entry.getId()
         val expectedParentId = currentEntryId.get()
 
-        if (entry.getParentId() != expectedParentId) {
+        val newEntry: HistoryEntry = if (swapParent) {
+            IntermediateHistoryEntry(entry.getContent(), expectedParentId)
+        } else {
+            entry
+        }
```
Davenury, line 36: I'm not sure about the consequences it will bring. Swaping this entry will result in having different entry id, but since we're adding it only now, id of this entry could even be null, so we shouldn't have relied on it, right?

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -1,433 +1,437 @@
...
+                true
+            )
+        }
+
+        signal(Signal.OnHandlingElectBegin, null, message.change)
+
+        transactionBlocker.assertICanSendElectedYou()
+
+        if (!this.checkBallotNumber(message.ballotNumber)) throw NotElectingYou(myBallotNumber, message.ballotNumber)
+        val initVal = if (!shouldCheckForCompatibility(message.change.peers) || history.isEntryCompatible(message.change.toHistoryEntry())) Accept.COMMIT else Accept.ABORT
```
Davenury, line 88: `My encoding is better than yours and to show my encoding is superior I will point out places I've changed in this file`
First one being here - I added `!shouldCheckForCompatibility`

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -1,433 +1,437 @@
...
+            if (isCurrentTransaction) {
+                this.transaction =
+                    this.transaction.copy(decision = true, acceptVal = Accept.COMMIT, ended = true)
+            }
+
+            if (message.acceptVal == Accept.COMMIT) {
+                signal(Signal.OnHandlingApplyCommitted, transaction, message.change)
+            }
+            if (message.acceptVal == Accept.COMMIT && !transactionWasAppliedBefore()) {
+                history.addEntry(message.change.toHistoryEntry(), !shouldCheckForCompatibility(message.change.peers))
```
Davenury, line 151: here, I'm passing `!shouldCheckForCompability`

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -1,433 +1,437 @@
...
+            logger.info("${getPeerName()} Releasing semaphore as cohort")
+            transactionBlocker.releaseBlock()
+
+            signal(Signal.OnHandlingApplyEnd, transaction, message.change)
+        }
+    }
+
+    private fun shouldCheckForCompatibility(peers: List<String>): Boolean {
+        return peers.size == 1
+    }
```
Davenury, line 165: I've also added this function

src/main/kotlin/com/github/davenury/ucac/history/History.kt
```diff
@@ -25,18 +25,24 @@ class History {
         )
     }
 
-    fun addEntry(entry: HistoryEntry) {
+    fun addEntry(entry: HistoryEntry, swapParent: Boolean = false) {
```
Rados13, line 28: Where do we use this function?

Rados13 (changes requested):
I don't understand this PR, maybe because of a lack of context.

src/main/kotlin/com/github/davenury/ucac/history/History.kt
Davenury: This function is used in handleApply in GPAC for sure, I don't know where else (probably consensus uses it too), that's why I added default value.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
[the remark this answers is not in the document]
Davenury: I'll answer here with the context of this PR: when we apply change now, we check if provided parentId is equal to the one that we expect (i.e. the ID of last change we did). All of our current tests (aside from this PR) execute only one change a top of the InitialHistoryChange which has the same ID in every peerset (since it's hash of {} string). Therefore no test has checked if our service works, if there are any other changes. So, for example:
Peerset1 has this state: I->ChangeA->ChangeB
And Peerset2 has this state:
I->ChangeC->ChangeD

Note that now, if we'd like to introduce change that should be present in both Peerset1 and Peerset2, we need to provide two parentIDs (which is final plan and according to Kamil we don't need that now). So, for now (and probably forever) we'll provide user request with only one parentID - the one of change from the peerset you're sending request to. Let's say (for example) that you (as user) send request for Peerset1, providing change with parentId = ChangeB.getId().
Now, when a leader from Peerset1 tries to execute this change, peers from Peerset2 checks their history and says that the change isn't compatible, as provided parentId isn't the one they expected (ChangeD.getId()), therefore they send initVal as ABORT. Leader sees that some of the peers sent ABORT, so he executes transaction to its end with according result (therefore not executing it at all). 
To prevent this behavior, I added a check if compability of changes should be checked - right now it's in trivial way - if change is executed in two peersets, changes shouldn't be validated. This allows peers from Peerset2 set their initVal to COMMIT.
The next change (one with swapping ids) is there to provide compability of parentIds - we use hashmap and iterate over ids, so we would have a break in chain of changes in Peerset2 if I didn't swap that id. As I wrote somewhere else, I don't know what exact consequences brings this swaping of id, but it's not any test-breaking behavior.

src/main/kotlin/com/github/davenury/ucac/history/History.kt
kjarosh: Instead of implementiclng hacks here, just call this function with different entry

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -160,6 +160,10 @@ class GPACProtocolImpl(
         }
     }
 
+    private fun shouldCheckForCompatibility(peers: List<String>): Boolean {
+        return peers.size == 1
```
kjarosh, line 164: This implementation requires a comment

kjarosh (changes requested):

Rados13 (dismissed):
Add comment required by Kamil and IMO PR is ready for merge.

PR #73, opened 2022-11-23
2PC

Rados13:


src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -56,8 +60,33 @@ class Worker(
     }
 }
 
+enum class ProcessorJobType {
```
Davenury, line 63: This is what I was talking about before and I think it could be somewhere 'higher'

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
 
 /**
  * @author Kamil Jarosz
  */
 interface AtomicCommitmentProtocol {
-    suspend fun proposeChangeAsync(change: Change): CompletableFuture<ChangeResult>
+
+    suspend fun performProtocol(change: Change)
+
+    fun getLogger(): Logger
```
Davenury, line 21: I don't think implementation of atomic commitment should be used to get logger, nor should it be used to get peer resolver
Edit after seeing the rest of the code: If you create abstract class from it, you can pass peerResolver and logger to constructor of this class.

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+
+    fun getPeerResolver(): PeerResolver
+
+    fun putChangeToCompletableFutureMap(change: Change, completableFuture: CompletableFuture<ChangeResult>)
```
Davenury, line 25: If this will be in all `AtomicCommitment` implementations, I'd change this interface to abstract class, have this map there and not create such method (since, I suppose, it will look the same across all implementations.

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+        val enrichedChange =
+            if (change.peers.contains(myAddress)) {
+                change
+            } else {
+                change.withAddress(myAddress)
+            }
+
+        putChangeToCompletableFutureMap(enrichedChange, cf)
+
+        GlobalScope.launch(MDCContext()) {
```
Davenury, line 42: GlobalScope? I think we should create executor or use one we've created. This could easily destabilize tests, since Global Scope won't be affected by shutting down application.

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+            performProtocol(enrichedChange)
+        }
+
+        return cf
+    }
+
+    fun getPeersFromChange(change: Change): List<List<PeerAddress>> {
+        if (change.peers.isEmpty()) throw RuntimeException("Change without peers")
```
Davenury, line 50: Since we're enriching the change with leader address, can something like that happen? If so, I'd use `IllegalStateException` but I guess it's just semantics, so you can ignore the exception change.

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+        val peerResolver = getPeerResolver()
+        return change.peers.map { peer ->
+            val peersetId = peerResolver.findPeersetWithPeer(peer)
+            if (peersetId == null) {
+                getLogger().error("Peer $peer not found in ${peerResolver.getPeers()}")
+            }
+            return@map peerResolver.getPeersFromPeerset(peersetId!!) // TODO we need to specify peersetIds instead of peers
```
Davenury, line 57: I think you can omit `return@map`

Davenury: Also this piece of code doesn't make much sense. If `peersetId` is null, executing `peersetId!!` will result in NullPointerException. You could get around that by executing something like this:
```kotlin
peerResolver.findPeersetWithPeer(peer)
    ?.let { peerResolver.getPeersFromPeerset(it) }
    ?: run { logger.error(...); throw SomeException() }
```
not sure about `run` though (but something like this should be ok)

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+        }.map { peerset -> peerset.filter { it.globalPeerId != peerResolver.currentPeerAddress().globalPeerId } }
```
Davenury, line 58: What does it do?

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+) : SignalSubject, AtomicCommitmentProtocol {
+
+    private var currentProcessedChange: TwoPCChange? = null
+    private var changeTimer: ProtocolTimer = ProtocolTimerImpl(twoPCConfig.changeDelay, Duration.ZERO, ctx)
+
+    private val mutex = Mutex()
+
+    private val changeIdToCompletableFuture: MutableMap<String, CompletableFuture<ChangeResult>> = mutableMapOf()
+
+    override fun getPeerName() = "peerset${myPeersetId}/peer${myNodeId}"
```
Davenury, line 37: Does this change between protocols? If not, I'd throw it to the interface / abstract class (whatever you'll decide)

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+
+    override suspend fun performProtocol(change: Change) {
+
+
```
Davenury, line 41: Too many enters (one is ok, two or more look weird 😄)

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+        val mainChangeId = change.toHistoryEntry().getId()
+
+        val acceptChange = TwoPCChange(change.parentId, TwoPCStatus.ACCEPTED, change.peers, change = change)
+
+        mutex.withLock {
+            currentProcessedChange = acceptChange
+        }
```
Davenury, line 48: Maybe that's because I'm stupid with concurrency but why do you have to lock only this line? I mean the moment ThreadA get out of this line, ThreadB can get in and override `currentProcessedChange`, which will result in overriding it in ThreadA. Or maybe I'm wrong 😅

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+
+        val acceptResult = checkChangeAndProposeToConsensus(acceptChange).await()
+
+        if (acceptResult.status != ChangeResult.Status.SUCCESS) changeConflict(
+            mainChangeId,
+            "failed during processing acceptChange"
```
Davenury, line 54: I'd mention that it's in 'TwoPC'

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+        logger.info("Decision $decision committed in all peersets $commitChange")
+        signal(Signal.TwoPCOnChangeApplied, change)
+
+        mutex.withLock {
+            currentProcessedChange = null
+        }
+
+        val result = if (decision) ChangeResult.Status.SUCCESS else ChangeResult.Status.CONFLICT
+        changeIdToCompletableFuture[mainChangeId]!!.complete(ChangeResult(result))
+    }
```
Davenury, line 90: I have a request - could you actally split it to multiple functions, each function representing a phase of 2PC? I think it would be more readable that way.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+
+
+    public suspend fun handleAccept(change: Change) {
+        if ((change is TwoPCChange).not()) {
```
Davenury, line 94: I think more readable would be `change !is TwoPCChange`

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+            logger.info("Received not 2PC change $change")
+            throw TwoPCHandleException("Received change of not TwoPCChange in handleAccept: $change")
+        }
+
+        mutex.withLock {
+            if (currentProcessedChange != null) throw TwoPCHandleException("Currently processing other 2PC change")
+        }
```
Davenury, line 101: I still think you aren't thread-safe.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+
+
+        val changeWithProperParentId = change.copyWithNewParentId(history.getCurrentEntry().getId())
+        val result = consensusProtocol.proposeChangeAsync(changeWithProperParentId).await()
+
+        if (result.status == ChangeResult.Status.SUCCESS) mutex.withLock {
+            currentProcessedChange = change as TwoPCChange
+        } else {
+            throw TwoPCHandleException("TwoPCChange doesn't applied change")
```
Davenury, line 110: `TwoPCChange didn't apply change` 😄

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        changeTimer.startCounting {
+            askForDecisionChange(change)
```
Davenury, line 114: I'd move the definition of the function right below this function.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    private suspend fun askForDecisionChange(change: Change, iteration: Int = 0) {
+        val resultChange = protocolClient.askForChangeStatus(
+            change.peers.first { it != myAddress() },
+            change
+        )
+
+        if (resultChange == null && iteration == twoPCConfig.maxChangeRetries)
+            throw TwoPCConflictException("We are blocked, because we didn't received decision change")
```
Davenury, line 191: We will be blocked forever in that case. And, as I understand, after maxChangeRetries we decide it's enough and if leader wants to continue with the transaction, he has to do the effort? But as we don't have the CC mechanizm and leader doesn't recover the change if he fails... we'll be blocked forever?

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+                    changeTimer.cancelCounting()
+                    checkChangeAndProposeToConsensus(change)
+                }
+
+                change.parentId == changeId &&
+                        change.copyWithNewParentId(currentProcessedChange!!.parentId) == currentProcessedChange!!.change -> {
+                    changeTimer.cancelCounting()
+                    val updatedChange = change.copyWithNewParentId(history.getCurrentEntry().getId())
+                    checkChangeAndProposeToConsensus(updatedChange)
+                }
```
Davenury, line 139: Did you include the test where we make change in both single peersets (different changes) and then introduce another change in both of there? This when is pretty complicated and the test I mentioned will make more of less sure that it works.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+        cf.await()
+    }
+
+    public suspend fun handleAskDecision(changeId: String): Change =
+        consensusProtocol.getState()
+            .toEntryList()
+            .find { it.getParentId() == changeId }
+            ?.let { Change.fromHistoryEntry(it) }
+            ?: throw ChangeDoesntExist(changeId)
```
Davenury, line 156: This function doesn't 'handle' anything - it just returns change, therefore I'd change it's name.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+            throw TwoPCConflictException("We are blocked, because we didn't received decision change")
+
+        if (resultChange != null) handleDecision(resultChange)
+        else changeTimer.startCounting {
+            askForDecisionChange(change, iteration + 1)
+        }
+    }
+
+    private fun checkChangeCompatibility(change: Change) =
+        if (history.isEntryCompatible(change.toHistoryEntry()).not()) {
```
Davenury, line 200: I don't feel the `.not()` at the end is more readable than `!` at the beginning. I feel like I ingest the whole expression and then there's "wait, it's actually negation of it all". But I guess it's a matter of taste.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+            logger.info("Change is not compatible with history $change")
+            throw HistoryCannotBeBuildException()
+        } else Unit
```
Davenury, line 203: do you have to include something like that? since the function doesn't return anything, imo using `=` syntax is pointless.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,232 @@
...
+
+    private suspend fun checkChangeAndProposeToConsensus(change: Change): CompletableFuture<ChangeResult> = change
+        .also { checkChangeCompatibility(it) }
+        .let { consensusProtocol.proposeChangeAsync(change) }
```
Davenury, line 207: `also` would be more syntactically correct, but `let` will also work.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPCProtocolClient.kt
```diff
@@ -0,0 +1,92 @@
...
+        return httpClient.post<Response>(url) {
+            contentType(ContentType.Application.Json)
+            accept(ContentType.Application.Json)
+            body = message!!
+        }
+    }
+
+
+    companion object {
+        private val logger = LoggerFactory.getLogger(TwoPCProtocolClientImpl::class.java)
```
Davenury, line 90: I think @kjarosh actually pointed out that it's better to use names, instead of classes. I don't know, so let him speak his mind.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPCProtocolClient.kt
```diff
@@ -0,0 +1,92 @@
...
+            }.let { coroutine ->
+                Pair(it.first, coroutine)
+            }
+        }.map {
+            val result = try {
+                it.second.await()
+                true
+            } catch (e: Exception) {
+                logger.error("$id - Error while evaluating response from ${it.first}: $e", e)
+                null
```
Davenury, line 69: Why `Boolean?`, since you don't return `false` in here?

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+        val signalListenersForCohort = mapOf(
+            Signal.ConsensusFollowerChangeAccepted to SignalListener {
+                println("Change accepted ${it.subject.getPeerName()} ${it.change}")
+                changeAppliedPhaser.arrive()
+            },
+            Signal.ConsensusLeaderElected to SignalListener { electionPhaser.arrive() }
+        )
+
+        val apps = TestApplicationSet(
+            listOf(3, 3),
```
Davenury, line 57: `changeAppliedPhaser` has 9 parties but you have 6 applications. How does it work XD?
Also, `electionPhaser` only has 5. Shouldn't we wait for all consensus peers to elect leader (if leader doesn't send this signal, ignore the second sentence).

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+                accept(ContentType.Application.Json)
+            }
+            fail("executing change didn't fail")
+        } catch (e: ClientRequestException) {
+            expectThat(e.message!!).contains("Change conflicted")
+        }
+        apps.stopApps()
+    }
+
+    @Disabled("Servers are not able to stop here")
```
Davenury, line 150: Are they? Wasn't this problem with ratis?

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+            appsToExclude = listOf(2, 5, 6, 7),
+        )
+        val peers = apps.getPeers()
+        val otherPeer = apps.getPeers()[1][0]
+
+        delay(5000)
+
+        // when - executing transaction
+        try {
+            executeChange("http://${peers[0][0]}/v2/change/sync?use_2pc", change(otherPeer))
```
Davenury, line 164: I think you should add `use_2pc=true` but I may be wrong

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+            val leaderElected = SignalListener {
+                electionPhaser.arrive()
+            }
+            listOf(
+                firstPeersetChangeAppliedPhaser,
+                secondPeersetChangeAppliedPhaser,
+                electionPhaser
+            ).forEach { it.register() }
+
+            var isChangeNotAccepted = true
```
Davenury, line 245: Shouldn't it be `AtomicBoolean` since we will use it in SignalListener?

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
```diff
@@ -0,0 +1,493 @@
...
+            askAllForChanges(peers.flatten()).forEach { changes ->
+                expectThat(changes.size).isEqualTo(2)
+                expectThat(changes[0]).isEqualTo(twoPCChange)
+                expectThat(changes[1]).isEqualTo(change.copyWithNewParentId(twoPCChange.toHistoryEntry().getId()))
+            }
+
+            apps.stopApps()
+        }
+
+    @Disabled("I am not sure if this is solid case in 2PC")
```
Davenury, line 297: imo it should be replace with test `transaction should be processed after leader fails and then starts up and wants to get the transaction to an end`

Davenury (changes requested):

modules/common/src/main/kotlin/com/github/davenury/common/Change.kt
```diff
@@ -65,6 +67,8 @@ sealed class Change {
     }
 
     abstract fun withAddress(myAddress: String): Change
+
+    abstract fun copyWithNewParentId(parentId: String): Change
```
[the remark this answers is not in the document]
Rados13: Nope, I can't because copy method is only available for the data class.

modules/common/src/main/kotlin/com/github/davenury/common/Change.kt
```diff
@@ -187,4 +202,47 @@ data class AddGroupChange(
...
+    override val parentId: String,
+    val twoPCStatus: TwoPCStatus,
+    override val peers: List<String>,
+    override val acceptNum: Int? = null,
+    val change: Change
+) : Change() {
+
+    override fun withAddress(myAddress: String): Change {
+        return this.copy(peers = peers.toMutableList().also { it.add(myAddress) })
+    }
```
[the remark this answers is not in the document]
Rados13: I can't because the `Change` class doesn't have access to copy method, because it is not a data class.

modules/common/src/main/kotlin/com/github/davenury/common/Change.kt
```diff
@@ -187,4 +202,47 @@ data class AddGroupChange(
...
+// TwoPC should always contain two changes:
+// If accepted: 2PCChange-Accept -> Change
+// Else: 2PCChange-Accept -> 2PCChange-Abort
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class TwoPCChange(
+    override val parentId: String,
+    val twoPCStatus: TwoPCStatus,
+    override val peers: List<String>,
+    override val acceptNum: Int? = null,
+    val change: Change
```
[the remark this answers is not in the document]
Rados13: Ok, it would imply that we will have to always pass named parameters for change and twoPCStatus or always explicitly pass null for `acceptNum`, but if you wish that. The customer is always right.

modules/common/src/main/kotlin/com/github/davenury/common/exceptions.kt
```diff
@@ -7,6 +7,9 @@ class NotValidLeader(val ballotNumber: Int, val messageBallotNumber: Int) : Exce
 class HistoryCannotBeBuildException : Exception()
 class AlreadyLockedException : Exception()
 class ChangeDoesntExist(changeId: String): Exception("Change with id: $changeId doesn't exists")
+class TwoPCConflictException(msg: String): Exception("During 2PC occurs error: $msg")
+class TwoPCHandleException(msg: String): Exception("In 2PC occurs error: $msg")
```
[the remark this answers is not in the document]
Rados13: My idea was that TwoPCConflictException suggests that some change ends with conflict and TwoPCHandleException is a situation the client sends a bad request to 2PC.

src/main/kotlin/com/github/davenury/ucac/ApplicationUcac.kt
```diff
@@ -80,6 +83,7 @@ class ApplicationUcac constructor(
     private val engine: NettyApplicationEngine
     private var raftNode: HistoryRatisNode? = null
     private var consensusProtocol: RaftConsensusProtocol? = null
+    private var twoPC: TwoPC? = null
```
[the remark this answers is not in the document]
Rados13: I don't think that we need another layer of abstraction here as we can't handle the current ones.

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
Rados13: Probably removing myself from the list of peers.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: Okay, you are right I should check first if any change is currently processed.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: Oh my god, I can't believe that someone decided to make something like `!is` instead of `is not`.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: We can talk about this on Monday.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: You know 2PC is blocking protocol so yes.

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: https://github.com/Davenury/Unifying-Consensus-And-Atomic-Commitment-PoC/blob/c980c11de190fe3ffce10a4c514a631ffdf72779/src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt#L350-L463

Yes because I copy all tests from GPAC and tried to adjust them for 2PC.

modules/common/src/main/kotlin/com/github/davenury/common/Change.kt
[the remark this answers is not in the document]
Davenury: As I mentioned - if it will introduce complications, drop it 😂

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: We return `completableFuture` from this function, so how I can use `also` here?

src/main/kotlin/com/github/davenury/ucac/ApplicationUcac.kt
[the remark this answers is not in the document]
Davenury: I actaully meant the interface Kamil made - there's the one that already is implemented by `TwoPC` and `GPAC`. In this interface we could do:
```
companion object {
    fun createInstance(config: Config): AtomicCommitment {
        if (config.comitment == "2PC") return TwoPC(config)
        return GPAC(config)
    }
}```
or so. But it assumes that we won't use both 2PC and GPAC together and I don't know if this assumption is valid.

File: src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt (hunk 13 above)
Comment (Davenury) [reply src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt:94]: XDDDDDD

File: src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt (hunk 14 above)
Comment (Davenury) [reply src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt:101]: I don't know if I will be remembering it on Monday :smile: neither if I have any more valuable input than being a vibe check "hmm, this still doesn't look thread-safe for me" :smile:

File: src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt (hunk 25 above)
Comment (Rados13) [reply src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt:57]: It's pretty simple the signal `ConsensusLeaderElected` is called when a peer receives a message  `LeaderElected` which is only sent to followers. As we are waiting that both peersets to elect leaders and each peerset has got 3 peers, so one peer has to be a leader and the other two peers are followers. So to sum up 2 peersets multiplied by 2 followers in each peerset give us 4 signals on which we are waiting for.

File: src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt (hunk 17 above)
Comment (Davenury) [reply src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt:191]: Yeah, of course I knew that :sweat_smile: I mean that it will be blocked forever in this case. shouldn't it ask forever? (so without the iteration condition?)

File: src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt (hunk 22 above)
Comment (Davenury) [reply src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt:207]: Ok, you can't :smile: I didn't notice we're returning it

File: src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt (hunk 25 above)
Comment (Rados13) [reply src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt:57]: As for `changeAppliedPhaser` it is similar because also only followers call that signal and each 2PC change consists of two consensus changes, so 2 multiplied by 4, gives us 8.

File: src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt (hunk 26 above)
Comment (Rados13) [reply src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt:150]: I am not sure but this test is disabled also in GPAC so I leave it as it is.

File: src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt (hunk 27 above)
Comment (Rados13) [reply src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt:164]: This test wasn't adjusted to 2PC as it is also disabled in GPAC.

File: src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt [hunk 35]
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+
+    fun getLogger(): Logger
+
+    fun getPeerResolver(): PeerResolver
+
+    fun putChangeToCompletableFutureMap(change: Change, completableFuture: CompletableFuture<ChangeResult>)
+
+    fun getChangeResult(changeId: String): CompletableFuture<ChangeResult>?
+
+    suspend fun proposeChangeAsync(change: Change): CompletableFuture<ChangeResult> {
```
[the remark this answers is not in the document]
Rados13: After Dawid's suggestion, I converted the interface to an abstract class.

src/main/kotlin/com/github/davenury/ucac/api/Worker.kt
```diff
@@ -56,8 +60,33 @@ class Worker(
...
-    val isConsensusOnly: Boolean = false
-)
+    val processorJobType: ProcessorJobType
+) {
+
+    constructor(
+        change: Change,
+        cf: CompletableFuture<ChangeResult>,
+        isOnePeersetChange: Boolean,
+        enforceGpac: Boolean,
```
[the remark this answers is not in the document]
Rados13: As always in this project.

src/main/kotlin/com/github/davenury/ucac/commitment/AtomicCommitmentProtocol.kt
```diff
@@ -2,11 +2,59 @@ package com.github.davenury.ucac.commitment
...
+        putChangeToCompletableFutureMap(enrichedChange, cf)
+
+        GlobalScope.launch(MDCContext()) {
+            performProtocol(enrichedChange)
+        }
+
+        return cf
+    }
+
+    fun getPeersFromChange(change: Change): List<List<PeerAddress>> {
```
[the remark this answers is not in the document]
Rados13: I moved all shared logic between 2PC and GPAC to AtomicCommitmentProtocol.

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
Rados13: As each peer stores the state in memory, so leader's failure will lead to the loss of his whole state. So it can't retry a transaction, also it will be very difficult to finish the transaction for the second peer if the first peer will be stopped or will be recovering from failure. But we can talk about this on chat or tomorrow.

src/main/kotlin/com/github/davenury/ucac/api/ProcessorJob.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.github.davenury.common.Change
+import com.github.davenury.common.ChangeResult
+import java.util.concurrent.CompletableFuture
+
+enum class ProcessorJobType {
+    CONSENSUS, GPAC, TWO_PC;
+
+    companion object {
+    }
```
Davenury, line 11: As the companion object is empty for now, I'd get rid of it

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,196 @@
...
+
+        val commitChange = if (decision) change
+        else TwoPCChange(
+            change.parentId, change.peers, change.acceptNum, twoPCStatus = TwoPCStatus.ABORTED, change = change
+        )
+//      Asynchronous commit change to consensuses
+        protocolClient.sendDecision(otherPeersets, commitChange)
+        val changeResult = checkChangeAndProposeToConsensus(commitChange.copyWithNewParentId(acceptChangeId)).await()
+
+        if (changeResult.status != ChangeResult.Status.SUCCESS) throw TwoPCConflictException("Change failed during committing locally")
```
Davenury, line 151: But this shouldn't happen, right? Consensus will be blocked from outside and will infinitely repeat this change?

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
```diff
@@ -0,0 +1,196 @@
...
+    private fun myAddress() = peerResolver.currentPeerAddress().address
+
+    private fun applySignal(signal: Signal, change: Change) {
+        try {
+            signal(signal, change)
+        } catch (e: java.lang.Exception) {
+            // TODO change approach to simulating errors in signal listeners
+            changeTimeout(change, e.toString())
+            throw e
+        }
```
Davenury, line 189: Singal is something we use for tests. Why the inability to execute signal should  timeout change in production code?

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocolAbstract.kt
```diff
@@ -4,44 +4,40 @@ import com.github.davenury.common.*
...
 import org.slf4j.LoggerFactory
 import java.lang.Exception
 import java.time.Duration
 import java.util.concurrent.CompletableFuture
 import kotlin.math.max
 
-interface GPACProtocol : SignalSubject, AtomicCommitmentProtocol {
-    override suspend fun proposeChangeAsync(change: Change): CompletableFuture<ChangeResult>
+abstract class GPACProtocolAbstract(peerResolver: PeerResolver, logger: Logger) : SignalSubject,
+    AbstractAtomicCommitmentProtocol(logger, peerResolver) {
```
Davenury, line 18: @kjarosh do we really need this interface (I know perfectly well I was the one who introduced it but I wasn't sober back then) - imo we don't

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocolAbstract.kt
```diff
@@ -4,44 +4,40 @@ import com.github.davenury.common.*
...
+    abstract suspend fun performProtocolAsLeader(change: Change, iteration: Int = 1)
+    abstract suspend fun performProtocolAsRecoveryLeader(change: Change, iteration: Int = 1)
+    abstract fun getTransaction(): Transaction
+    abstract fun getBallotNumber(): Int
 
 }
 
 
-class GPACProtocolImpl(
+class GPACProtocolImplAbstract(
```
Davenury, line 32: `GPACProtocolImpl` will suffice imo.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -419,23 +430,23 @@ class RaftConsensusProtocolImpl(
...
-                    logger.info("Response from leader: $response")
-                    response
-                } catch (e: Exception) {
-                    logger.info("Request to leader (${votedFor!!.address}) failed", e)
-                    null
-                } ?: ChangeResult(ChangeResult.Status.CONFLICT)
-                cf.complete(result)
-            }
+        changeIdToCompletableFuture[change.toHistoryEntry().getId()] = cf
+        GlobalScope.launch(MDCContext()) {
```
Davenury, line 434: usage of `GlobalScope` won't destabilize tests (even more)?

src/main/resources/application-integration.conf
```diff
@@ -9,6 +9,11 @@ ratis {
 
 peers = "127.0.0.1:8081,127.0.0.1:8082,127.0.0.1:8083;127.0.0.1:8084,127.0.0.1:8085,127.0.0.1:8086"
 
+twoPC {
+    changeDelay = PT5S
+    maxChangeRetries=5
+}
+
```
Davenury, line 16: Isn't it needed in other conf files?

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
Davenury: You said "pretty simple" but it's some high advanced math :smile:

src/main/kotlin/com/github/davenury/ucac/commitment/TwoPC/TwoPC.kt
Rados13: Kamil suggested that we should change it in the future so I add TODOs.

src/main/resources/application-integration.conf
Rados13: This param has got a default value and we only use it in 2PC tests so I don't see much benefit from adding it to other conf files, but as you requested that I will add this.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocolAbstract.kt
kjarosh: You mean `SignalSubject`?

Davenury (approved):

PR #74, opened 2022-11-24
Performance tests

Davenury:


.github/workflows/k8s.yml
```diff
@@ -2,21 +2,54 @@ name: K8s
...
-      - name: Build and publish image
-        uses: pmorelli92/github-container-registry-build-push@2.0.0
+#      - name: Build and publish image
+#        uses: pmorelli92/github-container-registry-build-push@2.0.0
+#        with:
+#          github-push-secret: ${{ secrets.DOCKER_DEPLOYMENT_TOKEN }}
+#          docker-image-name: ucac
+#          docker-image-tag: "${{ github.ref == 'refs/heads/main' && 'latest' || github.sha }}"
+#          dockerfile-path: ./Dockerfile.builder
+#          build-context: .
```
kjarosh, line 27: A comment to remove probably

cmd/commands/performance.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/spf13/cobra"
+
+	batchv1 "k8s.io/api/batch/v1"
+	"k8s.io/client-go/kubernetes"
+    v1 "k8s.io/api/core/v1"
+    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
```
kjarosh, line 12: Tabs or spaces?

Dockerfile
```diff
@@ -1,8 +1,19 @@
...
+###
+
+FROM adoptopenjdk:11-jre-hotspot as tests
+WORKDIR /tests
+
+ADD modules/tests/build/distributions/tests*.tar /tests
+
+EXPOSE 8080
+
+ENTRYPOINT ["bash", "-c", "/tests/tests*/bin/tests"]
```
kjarosh, line 19: These are really not stages, but rather different dockerfiles, and I would threat them as such

modules/tests/src/main/kotlin/com/github/davenury/tests/Chnages.kt
```diff
@@ -0,0 +1,68 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicReference
+
+/*
+* 1. Config for atomic commitment and consensus - which one to use
+* 2. Strategies for sending changes to peersets - what if there's no free peerset?
+* 3. Tests should have metrics too - e.g. how many requests per second are we sending
+* */
+
+class Changes(
```
kjarosh, line 18: There's a typo in the filename

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Routing.kt
```diff
@@ -62,11 +70,30 @@ fun Application.apiV2Routing(
...
 
+    suspend fun notify(notificationUrl: String?, change: Change, changeResult: ChangeResult) {
+        val decodedUrl = URLDecoder.decode(notificationUrl, Charset.defaultCharset())
+        try {
+            httpClient.post<HttpStatement>(decodedUrl) {
+                contentType(ContentType.Application.Json)
+                body = Notification(change, changeResult)
+            }
+        } catch (e: Exception) {
+            println("Couldn't send notification to $decodedUrl, ${e.message}")
```
kjarosh, line 81: `println`?

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Routing.kt
```diff
@@ -62,11 +70,30 @@ fun Application.apiV2Routing(
         }
     }
 
+    suspend fun notify(notificationUrl: String?, change: Change, changeResult: ChangeResult) {
```
kjarosh, line 73: I thought we want to minimize logic in routing?

Davenury: Yeah, we do - I totally forgot about it. Do you think it would be better to include this function directly in service or should I create class `Notifier` or so and include dependency to an interface in Service?

Dockerfile
Davenury: In this file I will agree, in `Dockerfile.builder` I'm staying as it is now, as both images originates from the same build

cmd/commands/performance.go
Davenury: Since it's Golang - who knows? I'm changing to tabs though

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
```diff
@@ -42,12 +47,39 @@ class ApiV2Service(
         ?: gpacProtocol.getChangeResult(changeId)
         ?: throw ChangeDoesntExist(changeId)
 
-    suspend fun addChange(change: Change, enforceGpac: Boolean = false): CompletableFuture<ChangeResult> =
+    suspend fun addChange(change: Change, enforceGpac: Boolean = false, notificationUrl: String? = null): CompletableFuture<ChangeResult> =
```
Rados13, line 50: Why do we need this notificationUrl?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM adoptopenjdk:11-jre-hotspot
+FROM adoptopenjdk:11-jre-hotspot as application
```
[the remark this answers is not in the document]
Davenury: I'm not master docker but I'll try to answer as well as I can.
You can name images that are built - look at the Dockerfile.builder, where the first image is tagged as builder. This not only allows you to refer to it later in the same dockerfile but also push it with option --target (just like we do it in flow). In this certain usage it is just a historical thing, that I added when in this file there was also dockerfile for tests. I forgot to delete it but I don't think it changes anything so imo it can stay (but I'm not master with docker, so I don't know for sure).

cmd/commands/performance/clean.go
```diff
@@ -0,0 +1,59 @@
...
+
+			clientset, err := utils.GetClientset()
+			if err != nil {
+				panic(err)
+			}
+
+			deleteJob(clientset)
+			deleteService(clientset)
+			deleteConfigmap(clientset)
+		},
```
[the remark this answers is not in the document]
Davenury: Yup, I think so :smile:

modules/tests/src/main/kotlin/com/github/davenury/tests/TestExecutor.kt
```diff
@@ -1,71 +1,68 @@
...
+            }
+        }
+
+        if (sentSinglePeersetChanges < numberOfRequestsToSendToSinglePeerset) {
+            sentSinglePeersetChanges++
+            return 1
+        }
+
+        return Random.nextInt(maxPeersetsInChange - 1) + 2
+    }
```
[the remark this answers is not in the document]
Davenury: You're not stupid. Upon deploying the test, you can pass how many requests you want to send to one peerset and how many requests you want to send to multiple peersets. This function determines if the request should be sent to single peerset or multiple peersets (if multiple peersets, then how much peersets). But there's a bug in here - we don't increment multiple peerset changes counter.

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Service.kt
Davenury: This is webhook type of notifications. When you execute async change, you can create an endpoint in your application that will receive notification when async operation will end. In this case, we use it for `delay_on_conflicts` strategy, where we don't want to send a second change to peersets when they didn't process the first one.

Rados13: Why don't we use sync endpoint with properly high timeout?

Davenury: I think that async and webhook is better because we won't starve threads - we will send many requests and, without this, lots of threads might be waiting for responses instead of just taking another request to send.

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -0,0 +1,74 @@
...
+        val senderAddress = peersAddresses.asSequence().shuffled().find { true }!!
+        val change = AddUserChange(getCurrentParentId(), "userName${counter}", otherPeers.asList())
+        return sender.executeChange(senderAddress, change)
+    }
+
+    fun getCurrentParentId(): String = parentId.get()
+
+    fun overrideParentId(newParentId: String) {
+        parentId.set(newParentId)
+    }
```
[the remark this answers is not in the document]
Davenury: It was easier for me to logically separate them (considering each of them has different parentId for changes).

Rados13 (dismissed):

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -0,0 +1,82 @@
...
+    private var counter = AtomicInteger(0)
+    private val handledChanges: MutableList<String> = mutableListOf()
+    private val mutex = Mutex()
+
+    suspend fun handleNotification(notification: Notification) {
+        logger.info("Handling notification: $notification")
+        if (!handledChanges.contains(notification.change.id)) {
+            (notification.change.peersets.map { it.peersetId }).forEach { peersetId ->
+                val parentId = notification.change.toHistoryEntry(peersetId).getId()
+                println("New parent id for $peersetId - $parentId")
```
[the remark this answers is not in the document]
Davenury: It was just for debugging

cmd/commands/whole.go
```diff
@@ -0,0 +1,136 @@
...
+	cmd.Flags().StringVar(&config.pushgatewayAddress, "pushgateway-address", "prometheus-prometheus-pushgateway.default:9091", "Pushgateway address")
+	cmd.Flags().BoolVar(&config.enforceAcUsage, "enforce-ac", false, "Determines if usage of AC protocol should be enforced even if it isn't required (GPAC)")
+	cmd.Flags().StringVar(&config.acProtocol, "ac-protocol", "gpac", "AC protocol to use in case it's needed. two_pc or gpac")
+	cmd.Flags().StringVar(&config.consensusProtocol, "consensus-protocol", "", "Consensus protocol to use. For now it's one protocol")
+	cmd.Flags().StringVar(&config.performanceTestTimeoutDeadline, "performance-test-timeout-deadline", "PT0S", "Additional duration after which test job should be force ended")
+
+	return cmd
+}
+
+func goBrr(config Config) {
```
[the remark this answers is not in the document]
Davenury: changed to `perform`

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -0,0 +1,82 @@
...
+
+class OnePeersetChanges(
+    private val peersAddresses: List<String>,
+    private val sender: Sender,
+    private val ownAddress: String
+) {
+    private var parentId = AtomicReference(InitialHistoryEntry.getId())
+
+    suspend fun introduceChange(change: AddUserChange): ChangeState {
+        val senderAddress = peersAddresses.asSequence().shuffled().find { true }!!
```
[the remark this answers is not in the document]
Davenury: Yeah, I don't know what I was thinking

Rados13 (approved):

PR #93, opened 2023-01-08
First version of transaction blocker

Rados13:


src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -157,10 +157,12 @@ class GPACProtocolImpl(
             } else {
                 changeSucceeded(message.change)
             }
+            transactionBlocker.releaseBlock()
         } finally {
             transaction = Transaction(myBallotNumber, Accept.ABORT, change = null)
 
-            transactionBlocker.releaseBlock()
+            if (transactionBlocker.isAcquired() && transactionBlocker.getProtocolName() == ProtocolName.GPAC)
+                transactionBlocker.releaseBlock()
```
kjarosh, line 165: Shouldn't this code execute only if protocol name is GPAC? Why aren't we checking it in other places?

src/main/kotlin/com/github/davenury/ucac/consensus/raft/domain/messages.kt
```diff
@@ -21,7 +21,9 @@ data class ConsensusHeartbeat(
 
 data class ConsensusHeartbeatResponse(
     val success: Boolean,
-    val term: Int
+    val term: Int,
+    val transactionBlocked: Boolean = false,
+    val incompatibleWithHistory: Boolean = false
```
kjarosh, line 26: Why do we need to inform the leader whether transaction is blocked? What does it mean that the changes are incompatible with history?

I would propose informing the leader:
1. which accepted changes were applied, and
2. whether the peer votes for a proposed change.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
Rados13: We need to inform leader that transaction is blocked to repeat changes in the next heartbeat. Informing that entry is incompatible is equal to voting against change.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/domain/messages.kt
Rados13: We need to inform leader that transaction is blocked to repeat changes in the next heartbeat. Informing that entry is incompatible is equal to voting against change

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
Rados13: Yup we probably should check it before each releasing transactionBlocker.

modules/common/src/main/kotlin/com/github/davenury/common/ProtocolName.kt
```diff
@@ -0,0 +1,5 @@
+package com.github.davenury.common
+
+enum class ProtocolName {
+    CONSENSUS, GPAC, TWO_PC;
+}
```
Davenury, line 5: Lack of new line

modules/common/src/main/kotlin/com/github/davenury/common/exceptions.kt
```diff
@@ -5,10 +5,13 @@ class UnknownOperationException(val desiredOperationName: String) : Exception()
 class NotElectingYou(val ballotNumber: Int, val messageBallotNumber: Int) : Exception()
 class NotValidLeader(val ballotNumber: Int, val messageBallotNumber: Int) : Exception()
 class HistoryCannotBeBuildException : Exception()
-class AlreadyLockedException : Exception()
-class ChangeDoesntExist(changeId: String): Exception("Change with id: $changeId doesn't exists")
-class TwoPCConflictException(msg: String): Exception("During 2PC occurs error: $msg")
-class TwoPCHandleException(msg: String): Exception("In 2PC occurs error: $msg")
+class AlreadyLockedException(protocol: ProtocolName) : Exception(
+    "We cannot perform your transaction, as another transaction is currently running with protocol ${protocol.name}"
```
Davenury, line 9: `protocol.name` will be uppercase - you could use `protocol.name.lowercase()`

Davenury (approved):

PR #96, opened 2023-01-09
Raft leader stops being leader if has got outdated history

Rados13:


Rados13:
These changes are included #98

PR #111, opened 2023-02-17
Rewrite votedFor and add leader propagation queue

kjarosh:


src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftLeader.kt
```diff
@@ -0,0 +1,122 @@
...
+        this.id = id
+        this.elected = false
+    }
+
+    suspend fun elect(id: GlobalPeerId) = mutex.withLock {
+        this.id = id
+        this.elected = true
+
+        propagateChanges(id)
+    }
```
Rados13, line 50: This function doesn't handle situations when you are elected as a leader. In that situation, you will block yourself from sending http requests to yourself and don't schedule sending heartbeats. Which then leads to very strange

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftLeader.kt
```diff
@@ -0,0 +1,122 @@
...
+    private fun propagateChanges(leaderId: GlobalPeerId) {
+        while (propagationRequests.isNotEmpty()) {
+            val req = propagationRequests.pop()
+            with(CoroutineScope(leaderRequestExecutorService)) {
+                launch(MDCContext()) {
+                    req.cf.complete(propagateChange(leaderId, req.change))
+                }
+            }
+        }
+    }
```
Rados13, line 91: Why does `propagateChange` call `propagateChanges` which calls another method `propagateChange`?

Rados13 (changes requested):
I have two issues. First I am not sure about renaming `VotedFor` to `RaftLeader`, because `votedFor` is a term from paper. Secondly, these changes make tests pass once on three times.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -692,54 +681,31 @@ class RaftConsensusProtocolImpl(
...
-//              TODO: Change after queue
-            else -> {
-                result.complete(ChangeResult(ChangeResult.Status.TIMEOUT, "There should always be a leader"))
-                throw Exception("There should always be a leader")
+        val electedLeader: GlobalPeerId?
+        mutex.withLock {
+            electedLeader = leader.elected()
+        }
+
+        if (electedLeader == globalPeerId) {
```
Davenury, line 697: As far as I understand, you can synchronize on L694 but when execution reaches L697 the electedLeader variable might be outdated?

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftLeader.kt
```diff
@@ -0,0 +1,122 @@
...
+                    "as it is not elected yet, queueing: $change")
+        }
+        return cf
+    }
+
+    private fun propagateChanges(leaderId: GlobalPeerId) {
+        while (propagationRequests.isNotEmpty()) {
+            val req = propagationRequests.pop()
+            with(CoroutineScope(leaderRequestExecutorService)) {
+                launch(MDCContext()) {
```
Davenury, line 86: why do you launch another MDCContext? shouldn't we use the one we have (somehow)?

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftLeader.kt
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+        }
+    }
+
+    private suspend fun propagateChange(
+        leaderId: GlobalPeerId,
+        change: Change,
+    ): ChangeResult {
+        logger.info("Propagating change to leader ($leaderId): $change")
```
Davenury, line 97: I don't think its intuitive that class named `RaftLeader` logs `propagating change to leader` which means that this is not raft leader.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftLeader.kt
```diff
@@ -0,0 +1,122 @@
...
+        val leaderAddress = peerResolver.resolve(leaderId)
+        return try {
+            val url = "http://${leaderAddress.address}/consensus/request_apply_change"
+            val response = httpClient.post<ChangeResult>(url) {
+                contentType(ContentType.Application.Json)
+                accept(ContentType.Application.Json)
+                body = change
+            }
+            logger.info("Response from leader: $response")
+            response
```
Davenury, line 107: I don't know if I'm right, but I'd advise not to call specific httpClient from here, which can make testing hard.

Rados13: We use specific httpClient from each module :upside_down_face:

Rados13 (commented):
This PR also looks good, maybe it's even more idiomatic for OOP. But I would also run it 50 times to check if there is no race condition.

PR #112, opened 2023-02-20
delay start of application for redis to stand up

Davenury:


cmd/commands/deploy.go
```diff
@@ -317,13 +317,18 @@ func createSingleContainer(containerName string, peerConfig utils.PeerConfig, im
...
 		Image:     imageName,
 		Resources: resources,
 		Args: []string{
 			"-Xmx500m",
 			"-Dcom.sun.management.jmxremote.port=9999",
 		},
+		Command: []string{
+			"sh",
+			"-c",
+			"sleep 2 && /application/bin/PoC -Xmx=500m",
```
kjarosh, line 330: I would add a healthcheck for redis inside the app along with timeout (configurable)

Davenury: In app? Or in docker image? Since we're relying heavily on redis in fault tolerance, I wouldn't like to write logic like "try to connect to redis three times and then use in memory history if you cannot" (especially since we're creating jedis instance in history class directly) - I'd rather use some bash script in the command, e.g. checking if redis responds to ping

Rados13 (approved):

PR #114, opened 2023-02-24
remove waiting for redis

Davenury:


Rados13 (dismissed):

Dockerfile
```diff
@@ -8,4 +8,4 @@ ADD build/distributions/PoC-*.tar /application
 
 EXPOSE 8080
 
-ENTRYPOINT ["sh", "-c", "while ! redis-cli ping; do sleep 1 ; done && /application/PoC-*/bin/PoC"]
+ENTRYPOINT ["sh", "-c", "if [ \"$config_persistence_type\" = \"REDIS\" ] ; then while ! redis-cli ping; do sleep 1 ; done ; fi && sh /application/bin/PoC"]
```
kjarosh, line 11: This setting does not have to be provided by an environment variable, I would strongly recommend to do this in code

Davenury: I tested it without providing the variable - the while is skipped.

modules/common/src/main/kotlin/com/github/davenury/common/history/InMemoryHistory.kt
```diff
@@ -27,35 +28,43 @@ class InMemoryHistory : History {
         )
     }
 
+    override fun isEntryCompatible(entry: HistoryEntry): Boolean =
+        meterRegistry.timer("history_is_entry_compatible").record<Boolean> {
```
kjarosh, line 32: Metrics shouldn't be gathered from existing implementations (SRP), you should create an implementation akin to CachedHistory for gathering metrics

kjarosh (changes requested):

modules/common/src/main/kotlin/com/github/davenury/common/history/InMemoryHistory.kt
Davenury: Imo they should - creating another class just for metrics is overkill and metrics should be gathered as close to the implemenation, especially if that's timer.

kjarosh (approved):

PR #115, opened 2023-02-24
Enable SinglePeersetApiSpec

kjarosh:


Rados13 (commented):
Looks good to me but I would run it on 50 repeats of CI tests to be sure that this doesn't lead to any race conditions.

.github/workflows/gradle.yml
```diff
@@ -4,6 +4,10 @@ on: [ push, pull_request ]
 
 jobs:
   build:
+    strategy:
+      matrix:
+        run0: [0, 1, 2, 3, 4]
+        run1: [0, 1, 2, 3, 4]
```
Rados13, line 10: What is it?

Rados13 (approved):

.github/workflows/gradle.yml
kjarosh: A temporary measure for executing tests multiple times

Dockerfile.testcontainers
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apk add --no-cache openjdk11
+
+WORKDIR /application
+
+ADD build/distributions/PoC-*.tar /application
+
+EXPOSE 8080
+
+ENTRYPOINT ["sh", "-c", "/application/PoC-*/bin/PoC"]
```
Davenury, line 11: imo there's no sense in creating another dockerfile solemnly for tests - what do we test in there? [Let's firstly merge this one](https://github.com/Davenury/Unifying-Consensus-And-Atomic-Commitment-PoC/pull/114) and then it will not be needed anymore

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -803,6 +807,22 @@ class RaftConsensusProtocolImpl(
 
     private fun getHeartbeatTimer() = ProtocolTimerImpl(heartbeatTimeout, heartbeatTimeout.dividedBy(2), ctx)
 
+    private suspend fun tryPropagatingChangesToLeader() {
+        // TODO mutex?
+        val votedFor = this.votedFor
+        if (votedFor == null || !votedFor.elected) return
+        while (true) {
+            val changeToBePropagated = changesToBePropagatedToLeader.poll() ?: break
```
Davenury, line 815: How will this work? The moment the queue is empty, the loop will end. What about next changes?

Davenury (changes requested):

Davenury (approved):

PR #117, opened 2023-02-26
Optimize raft

Rados13:


src/main/kotlin/com/github/davenury/ucac/consensus/raft/domain/Ledger.kt
```diff
@@ -55,7 +55,7 @@ data class Ledger(
         val newAcceptedItems = proposedEntries.take(index + 1)
 
         newAcceptedItems.forEach {
-            history.addEntry(it.entry)
+            if (!history.containsEntry(it.entry.getId())) history.addEntry(it.entry)
```
Davenury, line 58: imo it's not entirely thread safe here - but it's only my opinion and I'm not master of concurrent programming, so I may be wrong

Davenury (approved):

src/main/kotlin/com/github/davenury/ucac/consensus/raft/domain/Ledger.kt
Rados13: I am also not the master of it but we call `updateCommitIndex` always inside mutex, but I can add a comment about this.

PR #118, opened 2023-02-27
Sigterm

Davenury:


modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -44,11 +44,15 @@ class Changes(
...
-                logger.info("Introduced change $change to peersets with ids $ids with result: $result\n, entries ids will be: ${ids.map { it to change.toHistoryEntry(it).getId() }}")
-            } else {
-                logger.info("Change $change was rejected, freeing peersets $ids")
+            try {
+                val result = changes[ids[0]]!!.introduceChange(change)
+                if (result == ChangeState.ACCEPTED) {
+                    logger.info("Introduced change $change to peersets with ids $ids with result: $result\n, entries ids will be: ${ids.map { it to change.toHistoryEntry(it).getId() }}")
+                } else {
+                    logger.info("Change $change was rejected, freeing peersets $ids")
+                    getPeersStrategy.freePeersets(ids)
```
kjarosh, line 53: Is is needed here as it's executed in finally?

modules/tests/src/main/kotlin/com/github/davenury/tests/TestNotificationService.kt
```diff
@@ -79,11 +79,15 @@ class TestNotificationService {
             }
         }
 
-        GlobalScope.launch {
-            delay(2000)
-            testExecutor.startTest()
-            delay(2000)
-            closeService()
+        try {
```
kjarosh, line 82: Shouldn't this try be inside `launch`? You catch here exceptions thrown when launching and not thrown in launched code.

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
Davenury: Yeah, I had some massive lag - we cannot invoke `freePeersets` in finally - we don't want to do this in any case

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -44,9 +44,21 @@ class Changes(
 
             val change = createChangeStrategy.createChange(ids, changes)
 
-            val result = changes[ids[0]]!!.introduceChange(change)
+            val result = try {
+                changes[ids[0]]!!.introduceChange(change)
+            } catch (e: Exception) {
```
kjarosh, line 49: Should this exception be completely ignored?

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -44,9 +44,21 @@ class Changes(
...
+            }
             if (result == ChangeState.ACCEPTED) {
-                logger.info("Introduced change $change to peersets with ids $ids with result: $result\n, entries ids will be: ${ids.map { it to change.toHistoryEntry(it).getId() }}")
+                logger.info(
+                    "Introduced change $change to peersets with ids $ids with result: $result\n, entries ids will be: ${
+                        ids.map {
+                            it to change.toHistoryEntry(
+                                it
+                            ).getId()
+                        }
```
kjarosh, line 59: You could extract it to a variable

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
Davenury: nope, it should be at least logged :smile: (I'll still blame the brainlag though)

Davenury:
Changes and fixes implemented in another PR

PR #121, opened 2023-02-27
loki logs

Davenury:


modules/tests/src/main/resources/logback.xml
```diff
@@ -1,11 +1,26 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{YYYY-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+            <pattern>%d{YYYY-MM-dd HH:mm:ss.SSS} %-5level [%-11logger{36}] %msg \(%thread\)%n</pattern>
         </encoder>
     </appender>
+    <appender name="LOKI" class="com.github.loki4j.logback.Loki4jAppender">
+        <http>
+            <url>http://loki.ddebowski.svc.cluster.local:3100/loki/api/v1/push</url>
```
Davenury, line 9: Yes, I know but I have no idea how to make it better - maybe always deploy it to `loki` namespace but we don't have one on cyfronet cluster

kjarosh: I believe it can be configured by the application / syatem properties / env variables https://logback.qos.ch/manual/configuration.html#variableSubstitution

kjarosh (approved):

PR #123, opened 2023-02-28
Remove acceptNum

kjarosh:


src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -58,19 +58,20 @@ class GPACProtocolImpl(
...
+        // if (decision != null) {
+        //     // meaning that I'm the cohort that got apply for transaction of original leader
+        //     return ElectedYou(
+        //         message.ballotNumber,
+        //         Accept.COMMIT,
+        //         message.acceptNum,
+        //         Accept.COMMIT,
+        //         true
+        //     )
+        // }
```
Rados13, line 74: Do you know you remove part of the protocol with this one move? Mostly value discovery part.

kjarosh: Yes, but tests are not failing. Which means we have to cover this scenario in tests.

Davenury: Still, I'd leave it be until there is a test for that (especially, since we're about to do fault-tolerance tests) and this might be valid there.

PR #125, opened 2023-02-28
heapdump on oom

Davenury:


Davenury:
Out of memory is not gracefull killing of the pod - therefore neither JVM, nor `preStop` hook will have time to create heapdump - I won't close this PR yet, since the error we've experienced was just it - error, not oom, so there might be a chance that we'll be able to capture it, shall this error occur again.

PR #127, opened 2023-03-01
add peer labels to all resources

Davenury:


kjarosh (changes requested):
Is `cmd/heap.bin.hwcache` intentional?

Rados13 (approved):

PR #136, opened 2023-03-11
Add sendLeaderRequest retries

Rados13:


src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -752,19 +752,23 @@ class RaftConsensusProtocolImpl(
...
-                        }
-                    logger.info("Response from leader: $response")
-                    response
-                } catch (e: Exception) {
-                    logger.info("Request to leader (${votedFor!!.address}) failed", e)
-                    null
-                } ?: ChangeResult(ChangeResult.Status.TIMEOUT)
+                var result: ChangeResult? = null
+                while(result == null) {
+                    logger.info("Send request to leader again \n")
```
kjarosh, line 757:
```suggestion
                    logger.info("Send request to leader again")
```

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -305,7 +305,7 @@ class GPACProtocolImpl(
                 applySignal(Signal.BeforeSendingApply, this.transaction, change)
             } catch (e: Exception) {
                 transaction = Transaction(myBallotNumber, Accept.ABORT, change = null)
-                logger.info("performProtocolAsLeader releaseBlocker")
+                logger.info("Exception in Signal BeforeSendingApply")
```
Davenury, line 308: error + cause

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -752,19 +752,23 @@ class RaftConsensusProtocolImpl(
...
-                            body = change
-                        }
-                    logger.info("Response from leader: $response")
-                    response
-                } catch (e: Exception) {
-                    logger.info("Request to leader (${votedFor!!.address}) failed", e)
-                    null
-                } ?: ChangeResult(ChangeResult.Status.TIMEOUT)
+                var result: ChangeResult? = null
+                while(result == null) {
```
Davenury, line 756: This has a possibility to be infinite - either add a comment that it isn't (like after a period of time new leader will be chosen and this loop will break) or add some timeout and rejection of the change in case the timeout occurred (timeout can be a number of retries also)

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -752,19 +752,23 @@ class RaftConsensusProtocolImpl(
...
+                    logger.info("Send request to leader again")
+                    result = try {
+                        val response =
+                            httpClient.post<ChangeResult>("http://${votedFor!!.address}/consensus/request_apply_change") {
```
Davenury, line 760: I'd argue that it should also be in `RaftClient` or however this class is called, since here the implementation leaks to domain (what is S in SOLID) but I won't push it - it's long since this code was spaghetti, now it's more like Italian restaurant.

Davenury (approved):

kjarosh (approved):

PR #140, opened 2023-03-15
Half of accepts

Davenury:


src/main/kotlin/com/github/davenury/ucac/config.kt
```diff
@@ -1,77 +1,78 @@
...
+data class TwoPCConfig(val changeDelay: Duration = Duration.ofSeconds(120))
+
+data class GpacConfig(
+    val maxLeaderElectionTries: Int = 5,
+    val leaderFailDelay: Duration = Duration.ofSeconds(60),
+    val retriesBackoffTimeout: Duration = Duration.ofSeconds(120),
+    val initialRetriesDelay: Duration = Duration.ofSeconds(0),
+    val ftAgreeRepeatDelay: Duration = Duration.ofMillis(500),
+    val maxFTAgreeTries: Int = 5,
+    val abortOnElectMe: Boolean = false,
```
Rados13, line 56: Why do we need this?

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -491,23 +514,26 @@ class GPACProtocolImpl(
...
         return responses.withIndex()
             .all { (index, value) ->
                 val allPeers =
                     if (index == myPeersetId) peers[index].size + 1 else peers[index].size
                 val agreedPeers =
-                    if (index == myPeersetId) value.size + 1 else value.size
-                agreedPeers >= allPeers / 2F
-            } && allShards
+                    if (index == myPeersetId) {
+                        value.count { condition(it) } + 1
```
Rados13, line 531: I would change the `value` name to `responses` and `responsesPerShard` or something like this.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -491,23 +514,26 @@ class GPACProtocolImpl(
...
+                    } else {
+                        value.count { condition(it) }
+                    }
+                agreedPeers >= (floor(allPeers * 0.5) + 1)
```
Rados13, line 535: Also, I would create a private function `fun isMoreThanHalf(value) = floor(value*0.5)+1`.

src/main/kotlin/com/github/davenury/ucac/config.kt
Davenury: Tests - currently there's no easy way to force peer to vote `ABORT` since we only use history for that.

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -491,23 +514,26 @@ class GPACProtocolImpl(
     }
 
     private fun <T> superMajority(responses: List<List<T>>, peers: List<List<T>>): Boolean =
-        superFunction(responses, 2, peers)
+        ((floor(peers.size * 0.5) + 1) <= responses.size) && superFunction(responses, peers)
```
kjarosh, line 517: `superMajority` does not verify COMMIT / ABORT? (modeled as `condition`)

Davenury: I don't think so - we use superMajority only with recovery leader and if that's the case you already know the value, since that's the value you've received in fault tolerance phase

Rados13 (approved):

PR #143, opened 2023-03-18
Scrapper

Davenury:


misc/grafana-scrapping/index.js
```diff
@@ -0,0 +1,69 @@
...
+    panelId,
+    from,
+    to,
+    protocol,
+    experiment
+}) => {
+    const client = await page.target().createCDPSession()
+    await client.send('Page.setDownloadBehavior', {
+        behavior: 'allow',
+        downloadPath: `/home/dawid/workspace/studia/magisterka/results/20-03/${protocol}/${experiment}`
```
Rados13, line 25: It probably won't work on any other laptop than yours :laughing:

Rados13 (approved):

misc/grafana-scrapping/index.js
Davenury: meh, I saw that it creates the path, so I think it would - it would be hard to find though

misc/grafana-scrapping/index.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+(async () => {
+    const browser = await puppeteer.launch({headless: false});
+    const page = await browser.newPage();
+    await page.setViewport({ width: 1366, height: 768});
+
+    for (let experiment of experiments) {
+        for (let panel of panels) {
+            console.log(`panelId: ${panel}`)
```
Rados13, line 72: Do we need this?

Rados13 (approved):

misc/grafana-scrapping/index.js
Davenury: Big return of log hunter :smile: (no, we don't)

Rados13 (approved):

PR #149, opened 2023-03-20
do not await for jobs

Rados13:


src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -168,9 +175,14 @@ class GPACProtocolImpl(
         if (isCurrentTransaction) leaderFailTimeoutStop()
         signal(Signal.OnHandlingApplyBegin, transaction, message.change)
 
+        val entry = message.change.toHistoryEntry(globalPeerId.peersetId)
+
         when {
             !isCurrentTransaction && !transactionBlocker.isAcquired() -> {
-                transactionBlocker.tryToBlock(ProtocolName.GPAC, message.change.id)
+
+                if (history.containsEntry(entry.getId()))
```
kjarosh, line 183: Shouldn't a negation be there?

src/main/kotlin/com/github/davenury/ucac/ApplicationUcac.kt
```diff
@@ -222,11 +222,10 @@ class ApplicationUcac constructor(
             }
 
             exception<ChangeDoesntExist> { cause ->
+                logger.error("Change doesn't exist")
```
Davenury, line 225: I'd log the cause

src/main/kotlin/com/github/davenury/ucac/ApplicationUcac.kt
```diff
@@ -240,6 +239,7 @@ class ApplicationUcac constructor(
             }
 
             exception<GPACInstanceNotFoundException> { cause ->
+                logger.error("GPAC instance not found")
```
Davenury, line 242: here too

src/main/kotlin/com/github/davenury/ucac/commitment/gpac/GPACProtocol.kt
```diff
@@ -220,11 +227,16 @@ class GPACProtocolImpl(
                     state = changeResult.name.lowercase()
                 )
             }
-        } catch (e: Exception) {
-            transaction = Transaction(myBallotNumber, Accept.ABORT, change = null)
+        } catch (ex: Exception) {
+            logger.error("Exception during applying change, set it to abort", ex)
+            transaction =
+                transaction.copy(ballotNumber = myBallotNumber, decision = true, initVal = Accept.ABORT, change = null)
```
Davenury, line 233: I'd still argue that we should somehow handle such situation - we agreed to commit and now we cannot - I think we should try again. But it's not the case of that PR

Davenury (approved):

PR #153, opened 2023-03-24
Add PeerId & PeersetId

kjarosh:


modules/common/src/main/kotlin/com/github/davenury/common/PeersetId.kt
```diff
@@ -0,0 +1,23 @@
...
+    @JsonValue
+    override fun toString(): String = peersetId
+
+    // https://github.com/FasterXML/jackson-module-kotlin/issues/22
+    companion object {
+        @JvmStatic
+        @JsonCreator
+        fun create(string: String): PeersetId {
+            return PeersetId(string)
+        }
```
Davenury, line 21: I don't get it - serialization works with serialization of other data classes (e.g. GPAC messages) without such magic - I don't know why it wouldn't here.

Davenury (approved):

PR #141, opened 2023-03-17
Alvin implementation

Rados13:

PR #158, opened 2023-03-29
consensus affinity

Davenury:


src/test/kotlin/com/github/davenury/ucac/consensus/ConsensusAffinitySpec.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.ktor.client.request.*
+import kotlinx.coroutines.runBlocking
+import org.junit.jupiter.api.Test
+import strikt.api.expect
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+import strikt.assertions.isNotEqualTo
+import java.util.concurrent.Phaser
+
+class ConsensusAffinitySpec: IntegrationTestBase() {
```
Rados13, line 19: Why do you put it in a separate file, shouldn't it be with other consensus tests?

Rados13 (approved):

src/test/kotlin/com/github/davenury/ucac/consensus/ConsensusAffinitySpec.kt
Davenury: The class for consensus tests is long enough, this is somewhat related to consensus (that's why it's still in the package) but I didn't want to put in all in the same class.

Rados13: I know that the consensus tests class is pretty long, but in the future, this package will consist mostly of tests for different consensus protocols like Alvin and Paxos and the current ConsensusSpec will be renamed to RaftSpec and it will be a little bit misunderstanding that we have a class without specified protocol, so you can move it or I will move it in #141.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -96,7 +103,25 @@ class RaftConsensusProtocolImpl(
...
+                        logger.debug("Leader is alive, waiting for his messages")
+                        leaderAffinityWaiterTimer.startCounting { sendLeaderRequest() }
+                    } else {
+                        logger.debug("Leader is not alive, trying to become a leader")
+                        sendLeaderRequest()
+                    }
+                } else {
+                    logger.debug("I am affinity leader")
+                    sendLeaderRequest()
+                }
```
Rados13, line 122: In a situation like this `consensusAffinity.isEmpty() || consensusAffinity[peersetId] == null`, doesn't this behavior lead to that all peers will think that they are affinity leaders and will try to become a leader?

src/main/kotlin/com/github/davenury/ucac/common/PeersetProtocols.kt
```diff
@@ -33,6 +32,16 @@ class PeersetProtocols(
     val twoPC: TwoPC
     val gpacFactory: GPACFactory
 
+    fun beginConsensusProtocol() {
+        ctx.dispatch(Dispatchers.IO) {
```
Rados13, line 36: Why do we need this?

src/test/kotlin/com/github/davenury/ucac/consensus/ConsensusAffinitySpec.kt
Davenury: tbh I'd rather move both raft tests classes to package `consensus.raft` and create tests for other consensus protocols in `consensus.alvin` and `consensus.pigpaxos` packages but I won't push it

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -172,6 +197,12 @@ class RaftConsensusProtocolImpl(
                     "currentTerm=$currentTerm,lastApplied=${state.lastApplied}"
         )
 
+        if (affinityHandler.shouldRestartAffinityTimer(peerId)) {
+            timer.cancelCounting()
+            leaderAffinityWaiterTimer.cancelCounting()
+            leaderAffinityWaiterTimer.startCounting { sendLeaderRequest() }
+        }
```
Rados13, line 204: It looks very strange that we have to cancel two timers here, also we will again restart `timer` in `restartTimer()` function.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -884,7 +915,7 @@ class RaftConsensusProtocolImpl(
 
     private fun amILeader(): Boolean = role == RaftRole.Leader
 
-    private fun getHeartbeatTimer() = ProtocolTimerImpl(heartbeatTimeout, heartbeatTimeout.dividedBy(2), ctx)
+    private fun getHeartbeatTimer(offset: Duration = Duration.ZERO) = ProtocolTimerImpl(heartbeatTimeout.plus(offset), if (affinityHandler.amITrueAffinityLeader()) Duration.ZERO else heartbeatTimeout.dividedBy(2), ctx)
```
Rados13, line 918: Yyyy why do we need this, isn't affinityLeader something like only initial settings?

src/test/kotlin/com/github/davenury/ucac/consensus/ConsensusAffinitySpec.kt
Rados13: Is it some sort of convention in JVM or your private preference? Because I thought that this is pretty reasonable to have a package for consensus tests and in each file, we have tests for a specific protocol, similarly, we have tests in `api` package where we have one file for GPAC tests and another file for 2PC.

Davenury: I don't know about convention - at work we do it "by the sense of smell". There's also nothing that stops you from having nested packages, so you can have package `consensus` and then subpackage `consensus.raft` and there you could have a couple of files.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Davenury: And isn't that what we want? When there's no affinity, it's like normal consensus

src/main/kotlin/com/github/davenury/ucac/common/PeersetProtocols.kt
Davenury: `runBlocking` is blocking (no shit Sherlock, ik) but since we're relying here on logic that application runs, we cannot block the start of application (idk if that makes sense to you rn).
Basically previously we had "run blocking but it's just starts timer in another thread, so the application starts right off the bat", here the logic is a little bit more complicated, so we run the beginning of consensus protocol in another thread.

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Davenury: yeah, but that's because those timers do two different things, as far as I can tell. `timer` decides whether the leader should try to become leader once more and `leaderAffinityWaiterTimer` checks if the affinity leader didn't forget about him (e.g. died). `timer` has to be cancelled here because the affinity leader might not be the leader after the first iteration (e.g. only `peer17` voted for `peer3` to be affinity leader and `peer3` needs to request to be the leader once more and in such case, `peer17` would not get heartbeat and would try to be leader himself and that's not a situation we'd like)

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Davenury: Yeah, but you used this function also for `Candidate` and I didn't know if I didn't break something if I create other timer just for that. There's also a possibility that I'm destryoing the whole raft rn, that's why I pushed you for the review XD

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
Rados13: Not totally the same because in the current version on main, they will first wait one second plus random time from 0 to 2 seconds in your version everyone will start trying to become a leader at the same time.

PR #160, opened 2023-04-01
Clean up persistence (a step towards persistence for transaction blocker)

kjarosh:


modules/common/src/main/kotlin/com/github/davenury/common/persistence/JedisPersistence.kt
```diff
@@ -0,0 +1,49 @@
...
+        redis.call('SET', KEYS[1], ARGV[2])
+    end
+
+    return witness
+"""
+
+/**
+ * @author Kamil Jarosz
+ */
+class JedisPersistence(host: String, port: Int): Persistence {
```
Davenury, line 18: Jedis is a client - I'd rename it to "Redis"

src/main/kotlin/com/github/davenury/ucac/common/PersistenceFactory.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.github.davenury.common.persistence.JedisPersistence
+import com.github.davenury.common.persistence.Persistence
+import com.github.davenury.ucac.Config
+import com.github.davenury.ucac.PersistenceType
+
+/**
+ * @author Kamil Jarosz
+ */
+class PersistenceFactory {
+    fun createForConfig(config: Config): Persistence {
```
Davenury, line 14: Such method could be also done in interface - why create such class?

kjarosh: Due to dependency issues, `PersistenceFactory` is in a different module than `Persistence`.

modules/common/src/main/kotlin/com/github/davenury/common/persistence/JedisPersistence.kt
kjarosh: Renamed

Davenury (approved):

PR #161, opened 2023-04-02
always ask for consensus leader

Davenury:


modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -130,13 +132,18 @@ class OnePeersetChanges(
...
+        val address = peersAddresses.find { it.peerId == consensusLeaderId }!!
         return sender.executeChange(
-            senderAddress,
+            address,
             change
         )
     }
 
+    suspend fun getConsensusLeader(): PeerId =
+        sender.getConsensusLeaderId(peersAddresses.first())!!
```
Rados13, line 144: I think we could add TODOs for support with the use of Chaos-operator.

Rados13 (approved):

kjarosh (approved):

PR #166, opened 2023-04-03
try to fix traces in consensus

Davenury:


Rados13 (approved):
I assume that most of these changes are about traces. I don't know much about it, but I also don't see changes that could break protocols soo I approve it.

PR #170, opened 2023-04-06
peersets generation

Davenury:


misc/peersets-map-generator/map.py
```diff
@@ -0,0 +1,92 @@
...
+
+        chosen_peers = []
+
+        for i in range(number_of_peers_in_peerset):
+            random_value = abs(np.random.normal(loc=0, scale=self.skew))
+            current_list_size = self.number_of_peers - 1 -i
+            calculated_index = int(random_value * (self.number_of_peers - i))
+            correct_index = min(calculated_index, current_list_size)
+            chosen_peers.append(distances[correct_index][0])
+            del distances[correct_index]
```
Rados13, line 67: Could we move it to a function?

misc/peersets-map-generator/map.py
```diff
@@ -0,0 +1,92 @@
...
+        return chosen_peers
+
+    def _get_number_of_peers_in_peerset(self) -> int:
+        return random.randint(self.minimal, self.maximal)
+
+    @staticmethod
+    def _get_distance(peer_coords: Tuple[int, int], peerset_point_coords: Tuple[int, int]) -> float:
+        return math.sqrt(
+            math.pow(peer_coords[0] - peerset_point_coords[0], 2) + math.pow(peer_coords[1] - peerset_point_coords[1],
+                                                                             2))
```
Rados13, line 78: What do you think about this?
```suggestion
        return ((peer_coords[0] - peerset_point_coords[0])**2) + (peer_coords[1] - peerset_point_coords[1])**2)**(0.5)
```
Or this?
```suggestion
distance = lambda pointA,pointB, param: math.pow(pointA[param] - pointB[param], 2)
        return math.sqrt(distance(peer_coords, peerset_point_cords, 0) + distance(peer_coords, peerset_point_cords, 1))
```
Or the last one
```suggestion
distance = lambda pointA,pointB, param: math.pow(pointA[param] - pointB[param], 2)
        return math.sqrt(sum([distance(peer_coords, peerset_point_cords, param) for param in [0,1]]))
```

misc/peersets-map-generator/map.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @staticmethod
+    def _sort_distances(distances: Dict[str, float]) -> Tuple[str, float]:
+        # Note - works only with python 3.7+
+        return sorted(distances.items(), key=lambda item: item[1])
+
+    @staticmethod
+    def _get_random_coordinates():
+        return random.randint(0, 50), random.randint(0, 50)
```
Rados13, line 87: Shouldn't this function be parameterized?

Rados13 (approved):

misc/peersets-map-generator/map.py
```diff
@@ -0,0 +1,92 @@
...
+    @staticmethod
+    def _get_distance(peer_coords: Tuple[int, int], peerset_point_coords: Tuple[int, int]) -> float:
+        return math.sqrt(
+            math.pow(peer_coords[0] - peerset_point_coords[0], 2) + math.pow(peer_coords[1] - peerset_point_coords[1],
+                                                                             2))
+
+    @staticmethod
+    def _sort_distances(distances: Dict[str, float]) -> Tuple[str, float]:
+        # Note - works only with python 3.7+
+        return sorted(distances.items(), key=lambda item: item[1])
```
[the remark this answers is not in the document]
Davenury: I think I thought it would be much more complex. Even though it turned out to be not, I'd leave it as it is, since `sort_distances` is imo more readable

misc/peersets-map-generator/map.py
Davenury: The forbidden word? I don't think parametrization is necessary here, since the size of the map does not change anything (I think) in the algorithm of finding the peers but I'll parametrize it.

PR #171, opened 2023-04-08
increase grpc

Davenury:
Solves errors like `
ResourceExhausted desc = trying to send message larger than max (6365334 vs. 4194304)`


Rados13 (approved):
Could we disable temporarily this failing test in `SinglePeersetSpec`, if we don't plan to fix it in near future?

PR #173, opened 2023-04-10
ask for leader

Davenury:


modules/tests/src/test/kotlin/com/github/davenury/test/ChangesTest.kt
```diff
@@ -201,7 +204,8 @@ class ChangesTest {
             peers, DummySender(shouldNotify = false, peers), strategy, DefaultChangeStrategy(
                 ownAddress
             ),
-            null
+            null,
+            ""
```
Rados13, line 208: Could we make it default?

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -18,32 +21,48 @@ class Changes(
...
             }
+        }
+    }
 
-            it.map { pair -> pair.first }.zip(results).map { (peersetId, peerId) ->
-                peersetId to peers[peersetId]!!.find { peerAddress -> peerAddress.peerId == peerId }!!
+    private fun subscribeToStructuralChanges() {
+        GlobalScope.launch {
+            peers.entries.forEach { (_, addresses) ->
+                addresses.forEach { address ->
```
Rados13, line 55: Why didn't you use flatMap?

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -145,24 +167,67 @@ class OnePeersetChanges(
...
+        consensusLeader.set(getConsensusLeader())
+    }
+
+    private suspend fun getConsensusLeader(peerAddresses: List<PeerAddress> = peersAddresses): PeerAddress {
+        val address = peersAddresses.firstOrNull() ?: throw IllegalStateException("I have no more peers to ask!")
+        val peerId = try {
+            sender.getConsensusLeaderId(address)
+        } catch (e: Exception) {
+            logger.info("$address is dead, I'm trying to get consensus leader from another one")
+            return getConsensusLeader(peerAddresses.filterNot { it == address })
```
Rados13, line 210: I think this will be more efficient
```suggestion
            return getConsensusLeader(peerAddresses.drop(1))
```

Rados13 (approved):

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
Davenury: One must be aware how to use flatMap to use flatMap :smile:

modules/common/src/main/kotlin/com/github/davenury/common/ChangeResult.kt
```diff
@@ -2,7 +2,8 @@ package com.github.davenury.common
 
 data class ChangeResult(
     val status: Status,
-    val detailedMessage: String? = null
+    val detailedMessage: String? = null,
+    val desiredParentId: String? = null,
```
Rados13, line 6: Why do we need that?

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
Rados13: So it's a great occasion to learn `flatMap`.

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -145,24 +169,67 @@ class OnePeersetChanges(
...
+        return try {
+            httpClient.get("http://${consensusLeader.get()!!.address}/v2/last-change")
+        } catch (e: Exception) {
+            when (e) {
+                is ClientRequestException, is ServerResponseException -> throw e
+                else -> {
+                    logger.info("Consensus leader ${consensusLeader.get()} is dead, I'm trying to get a new one")
+                    populateConsensusLeader()
+                    return getChange()
+                }
```
Rados13, line 197: How does it work? Because all responses with status codes 400 and 500 will go to the first case, so to the second will go only 300?

Rados13 (approved):

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
Davenury: 300 is not an error - it's redirect (BTW)
other Exceptions might be from OKHTTP (e.g. connection refused)

modules/common/src/main/kotlin/com/github/davenury/common/ChangeResult.kt
Davenury: Since every peer can fault at any time, a notification might not get to the performance test job. We wouldn't know if that occurred, so we send desiredParentId on rejected status, so we can override it in performance test, so the test might continue (we'll check the history and the changes that were sent to tests later, so we'll check it)

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
Rados13: I found [that](https://github.com/ktorio/ktor/blob/68f1260e6814c9464e9f9d9e763d0d4cc1f36203/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultResponseValidation.kt#L53) so I thought ktor will throw exception also on redirects.

modules/common/src/main/kotlin/com/github/davenury/common/ChangeResult.kt
kjarosh: `currentEntryId`?

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
kjarosh: I would catch `IOException` here

modules/tests/src/main/kotlin/com/github/davenury/tests/Changes.kt
```diff
@@ -145,24 +169,67 @@ class OnePeersetChanges(
...
-            .also { this.consensusLeader = it }
+    suspend fun populateConsensusLeader() {
+        consensusLeader.set(getConsensusLeader())
+    }
+
+    private suspend fun getConsensusLeader(peerAddresses: List<PeerAddress> = peersAddresses): PeerAddress {
+        val address = peersAddresses.firstOrNull() ?: throw IllegalStateException("I have no more peers to ask!")
+        val peerId = try {
+            sender.getConsensusLeaderId(address)
+        } catch (e: Exception) {
```
kjarosh, line 210: `IOException`?

src/main/kotlin/com/github/davenury/ucac/api/ApiV2Routing.kt
```diff
@@ -125,5 +128,11 @@ fun Application.apiV2Routing(
         get("/v2/last-change") {
             call.respond(service.getLastChange() ?: HttpStatusCode.NotFound)
         }
+
+        post("/v2/subscribe-to-structural-changes") {
```
kjarosh, line 132: I would associate the word `structural` with organizational structures we are operating on, maybe a different name?

src/main/kotlin/com/github/davenury/ucac/common/structure/HttpSubscriber.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.github.davenury.common.PeerId
+import com.github.davenury.common.PeersetId
+import com.github.davenury.ucac.httpClient
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import io.ktor.http.*
+import org.slf4j.LoggerFactory
+
+class HttpSubscriber(
+    val address: String
```
kjarosh, line 13: `url`?

src/main/kotlin/com/github/davenury/ucac/common/structure/Subscribers.kt
```diff
@@ -0,0 +1,23 @@
+package com.github.davenury.ucac.common.structure
+
+import com.github.davenury.common.PeerId
+import com.github.davenury.common.PeersetId
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+
+object Subscribers {
```
kjarosh, line 9: I would refactor this into a class

modules/tests/src/test/kotlin/com/github/davenury/test/ChangesTest.kt
Davenury: Well we do not actually want to make it as default "" - in actual code it would not make any sense and making it default just for tests is imo not a way to do things.

src/main/kotlin/com/github/davenury/ucac/common/structure/Subscribers.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun notifyAboutConsensusLeaderChange(newPeerId: PeerId, newPeersetId: PeersetId) {
+        GlobalScope.launch {
+            subscribers.forEach {
+                launch {
+                    it.notifyConsensusLeaderChange(newPeerId, newPeersetId)
+                }
+            }
+        }
+    }
```
Rados13, line 24: Do we need this outside launch?

Rados13 (approved):

src/main/kotlin/com/github/davenury/ucac/common/structure/Subscribers.kt
Davenury: yes, notice that this function is not suspend, we need to launch it in a coruotine scope (I'm starting to rethink my life choices and usage of GlobalScope here though). Then each subscriber is notified in it's own launch.

Rados13 (approved):

PR #174, opened 2023-04-12
Two pc ft fix

Davenury:


Davenury:
TODO - add reaction to changes of consensus leaders

misc/peersets-map-generator/map.py
Rados13, line 1: Wasn't it added in a different PR?

modules/common/src/main/kotlin/com/github/davenury/common/ChangeResult.kt
```diff
@@ -31,6 +32,11 @@ data class ChangeResult(
         /**
          * Change was applied with ABORT result
          */
-        ABORTED;
+        ABORTED,
+
+        /**
+         * Change was not applied because I'm not the consensus leader
+         */
+        REDIRECT;
```
Rados13, line 40: Why do we need this, if I remember correctly, if we are not a leader we should forward this change to a leader.

cmd/chaos/ddebowski.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: lsc.davenury.github.com/v1
+kind: Scenario
+metadata:
+  name: gpac-chaos-ddebowski
+  namespace: ddebowski
+spec:
+  phases:
+    - duration: PT30S
```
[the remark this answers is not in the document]
Davenury: What kind of description do you have in mind? Like how to create one, what does each firld mean or more like "this scenario does this and this"

modules/common/src/main/kotlin/com/github/davenury/common/ChangeResult.kt
Davenury: In consensus yes. In twoPC we talked about it on Monday and decided that twoPC leader should only communicate with other consensus leader - therefore if the peer is not a consensus leader, it will be redirected to consensus leader.

src/test/kotlin/com/github/davenury/ucac/api/TwoPCSpec.kt
```diff
@@ -1,676 +1,867 @@
...
+            .filterNot { it == firstConsensusLeader }
+            .forEach {
+                val change = askForChanges(apps.getPeer(it)).last()
+                expectThat(change).isA<TwoPCChange>()
+                expectThat((change as TwoPCChange).twoPCStatus).isEqualTo(TwoPCStatus.ABORTED)
+            }
+    }
+
+    @Test
+    @Disabled("Sometimes consensus adds entry to history but doesn't send signal - bother Radek about it")
```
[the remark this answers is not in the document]
Davenury: I noticed that this test didn't want to end because some peers didn't send a signal even though entry was added to the history - as you can see I was supposed to bother you XD

src/main/kotlin/com/github/davenury/ucac/consensus/raft/infrastructure/RaftConsensusProtocolImpl.kt
```diff
@@ -824,6 +846,16 @@ class RaftConsensusProtocolImpl(
...
 
+            redirectIfNotLeader && votedFor?.elected == true -> {
+                logger.info("I'm no leader but there's a leader, so I'm sending redirect")
+                result.complete(ChangeResult(ChangeResult.Status.REDIRECT, currentConsensusLeader = votedFor!!.id))
+            }
+
+            redirectIfNotLeader && votedFor?.elected != true -> {
+                logger.info("I'm no leader but the leader has not been chosen yet, so I'm going to reject the change")
+                result.complete(ChangeResult(ChangeResult.Status.REJECTED))
+            }
```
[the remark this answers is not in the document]
Rados13: If this is related to 2PC I don't think that this should be done in the Raft module.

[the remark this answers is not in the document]
Davenury: I don't think there is other way to do that - raft is fixed on pushing the change forward. There are three reasons I decided to modify it like this:
* I recall you've said that originally in Raft behavior implemented here is desired
* Notice that it is only possibility in Raft, not a permament change in Raft behavior
* Yes, this change is introduced beacuse of 2PC but is purely Raft-related - we're still clear (this functonality may not only be used by 2PC but is provided by Raft)

src/main/kotlin/com/github/davenury/ucac/commitment/twopc/TwoPC.kt
```diff
@@ -376,10 +378,9 @@ class TwoPC(
     private suspend fun checkChangeAndProposeToConsensus(
         change: Change,
         originalChangeId: String,
-        redirectIfNotConsensusLeader: Boolean = false
     ): CompletableFuture<ChangeResult> = change
         .also { checkChangeCompatibility(it, originalChangeId) }
-        .let { consensusProtocol.proposeChangeAsync(change, redirectIfNotConsensusLeader) }
+        .let { consensusProtocol.proposeChangeAsync(change) }
```
Rados13, line 383: Don't we require verification if we are a leader also here?

Rados13 (dismissed):

src/main/kotlin/com/github/davenury/ucac/commitment/twopc/TwoPC.kt
Davenury: Those are used either in the same peer (in propose and decision phase) or within decision phase. The first one is assured by performance tests (they always send to consensus leader), in the second I think we should not, because when it isn't a consensus leader, twoPC peer will just ignore the message (curiosity of asynchronously making decision), so redirect here would cause the change to be asked only within timeout

src/test/kotlin/com/github/davenury/ucac/utils/ApplicationTestcontainersEnvironment.kt
```diff
@@ -1,6 +1,7 @@
 package com.github.davenury.ucac.utils
 
 import com.github.davenury.common.PeerId
+import com.github.davenury.ucac.common.structure.Subscribers
```
kjarosh, line 4: unused?

Rados13 (dismissed):

Rados13 (approved):

Rados13 (approved):

PR #175, opened 2023-04-17
Results persistence

Davenury:


cmd/commands/persistence/persist.go
```diff
@@ -0,0 +1,106 @@
...
+const (
+	YYYYMMDD = "2006-01-02"
+)
+
+func getCurrentPath() string {
+	path, err := os.Getwd()
+	if err != nil {
+		log.Println(err)
+	}
+	return path
```
Rados13, line 29: This way won't we return the path, even though an error occurs during getting it?

Rados13 (approved):

cmd/commands/persistence/persist.go
```diff
@@ -0,0 +1,106 @@
...
+
+type FileInfo struct {
+	Pod     string
+	SrcPath string
+	DstPath string
+	Name    string
+}
+
+const (
+	YYYYMMDD = "2006-01-02"
```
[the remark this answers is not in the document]
Davenury: That was the response from stack overflow. I guess that it works only for format (like YYYY-MM-DD in JVM, but it's golang so...)

cmd/commands/persistence/persist.go
Davenury: Yeah... we do... :_:

Rados13 (approved):

PR #181, opened 2023-04-28
Multiple peersets command

Davenury:


cmd/commands/deploy.go
```diff
@@ -17,6 +17,8 @@ import (
 
 type DeployConfig struct {
 	NumberOfPeersInPeersets []int
+	NumberOfPeersV2         int
+	PeersetsConfigurationV2 string
```
Rados13, line 21: Why do they have V2 in their name?

Rados13 (approved):

cmd/commands/deploy.go
Davenury: Backwards compability - this is far different from the new way of doing things, yet I didn't want to get rid of the possibility to deploy tests the old way.

kjarosh (approved):

